PR #1, opened 2019-05-17
Add Caddy

mholt:
I think the only thing missing is I'm not sure what to do when Intermediate or Old are selected. Caddy defaults to TLS 1.3 which (in Go at least) doesn't allow customizing cipher suites. Mozilla's current recommendation for Modern says only TLSv1.2, but I'm assuming that just needs to be updated.

In any case, here is how to get a rough equivalent of Modern in Caddy. It's at least a start!

### DIFF
--- a/src/js/configs.js
+++ b/src/js/configs.js
@@ -19,6 +19,13 @@ module.exports = {
     supportsOcspStapling: false,
     usesOpenssl: false,
   },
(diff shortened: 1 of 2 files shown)


april:
I'll make a few tweaks but overall this looks great. Thanks @mholt.  :)

PR #7, opened 2019-05-26
Traefik Configuration Support

captn3m0:
I've added basic support for TLS Versions, and HSTS.

Need suggestions on how to approach the Cipher Suite list. Traefik uses https://godoc.org/crypto/tls#pkg-constants, which is not used by any other configuration yet.

### DIFF
--- a/src/templates/partials/traefik.hbs
+++ b/src/templates/partials/traefik.hbs
@@ -1,13 +1,9 @@
 # generated with: {{{output.link}}}
 [entryPoints]
 [entryPoints.http.tls]
-{{#if (includes "modern" form.config)}}
-minVersion = "VersionTLS12"
-{{else if (includes "intermedia" form.config)}}
-minVersion = "VersionTLS10"
-{{else}}
-minVersion = "VersionSSL30"
-{{/if}}
+
+{{! See https://regexr.com/4f3e4 for an explanation of the regex. }}
+minVersion = "{{replace (concat (last output.protocols) "0") (toregex "([A-Z]{3})v(\d)\.?(\d)0?") "Version$1$2$3" }}"
 
 cipherSuites = [
 	"TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
--- a/src/templates/partials/traefik.hbs
+++ b/src/templates/partials/traefik.hbs
@@ -1,0 +1,36 @@
+# generated with: {{{output.link}}}
+[entryPoints]
+[entryPoints.http.tls]
+{{#if (includes "modern" form.config)}}
+minVersion = "VersionTLS12"
+{{else if (includes "intermedia" form.config)}}
+minVersion = "VersionTLS10"
+{{else}}
+minVersion = "VersionSSL30"
+{{/if}}
+
+cipherSuites = [
+	"TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
+	"TLS_RSA_WITH_AES_256_GCM_SHA384"
+]
+
+{{#if form.hsts}}
+# Traefik configuration for HSTS is on a per-host basis,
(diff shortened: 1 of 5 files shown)


april:
This PR has a bunch of stuff not related to traefix, including the reformatting of `config.js`, fixes to comments, and the like. I'll open up review for the related stuff however.  :)

src/templates/partials/traefik.hbs
```diff
@@ -0,0 +1,36 @@
+# generated with: {{{output.link}}}
+[entryPoints]
+[entryPoints.http.tls]
+{{#if (includes "modern" form.config)}}
```
april: This is not the way to do this, it doesn't update as time goes along. The template has access to `output.protocols`, which is an array of the supported protocol versions. You can use the last element to get the minimum version, and then use the `replace` helper to change `TLSv1.2` to `VersionTLS12`, for example.

src/templates/partials/traefik.hbs
```diff
@@ -0,0 +1,36 @@
...
+minVersion = "VersionTLS12"
+{{else if (includes "intermedia" form.config)}}
+minVersion = "VersionTLS10"
+{{else}}
+minVersion = "VersionSSL30"
+{{/if}}
+
+cipherSuites = [
```
april: You'll want to use a loop over `output.cipherSuites` to list your cipher suites, instead of hard codingf them. Take a look at how `awselb.hbs` does it to get a good idea of how it works.  :)

captn3m0:
Thanks for the review. The extra changes were due to my auto-formatter, will remove them from the branch.

Will get back to this in a few days.

src/templates/partials/traefik.hbs
captn3m0: A simple loop doesn't work because the golang constants that Traefik uses are very different from what other templates use so far.

https://godoc.org/crypto/tls#pkg-constants

The internet tells me to check the openssl constants and use them: https://www.woolie.co.uk/article/mapping-openssl-ciphers-to-iana-cipher-suite-registy/, but those aren't exposed currently.

Any suggestions on what route to take here?

captn3m0: @april Any suggestions?

april:
BTW, my plan is to (eventually) update the config system such that it contains both the OpenSSL cipher names and the IANA cipher names. That will hopefully make your life a lot easier.  :)

april:
See also: https://github.com/mozilla/ssl-config-generator/issues/21

april:
I swear I'm not ignoring you, I'm just swamped at the moment.  :)

dalanmiller:
@april once again, this is so great and something I wish I had when starting out with nginx and others. Thanks for your tweet directing here! 

Just as a heads up, this effort may want to wait until v2.0 is released.

There's still quite a bit to do[0] but if returning to the Traefik config after release isn't something there's appetite / scheduled time for, then waiting might be the best approach. 

@ldez -- are there release plans as to when the config changes for v2.0 will be solidified? 

[0] - https://github.com/containous/traefik/projects/3

ldez:
Traefik v2 is not ready for now, then have a Traefik v1 configuration is a good idea.

When the v2 will be release I hope you (@dalanmiller) will create a PR on this repository to update the configuration :wink: 

Thanks to @captn3m0 to spend time to do that PR :+1:

april:
So I used a ton of this work and pushed up a beta of Traefik support yesterday. Could an expert take a look at it for me, please?

Also, I still wasn't 100% certain how to add the HSTS headers when using the `entryPoints` thing, from the Traefik docs. Thanks for everybody's help and patience on this!

april:
Bump.  :)

captn3m0:
I'll file another PR with just the header stuff, since most of this is already covered.

PR #15, opened 2019-06-13
official branding "moz://a" is lowercase

mozfreddyb:
Changing "M" into "m"


mozfreddyb:
:blush:

april:
Now this is the kind of pull request I can get behind!  :)

PR #30, opened 2019-07-02
Fix duplicate protocol for caddy/modern

angristan:
Unless I'm missing something 🙂

### DIFF
--- a/src/templates/partials/caddy.hbs
+++ b/src/templates/partials/caddy.hbs
@@ -24,7 +24,7 @@ tls {
 {{#if (includes "modern" form.config)}}
 
 tls {
-    protocols tls1.3 tls1.3
+    protocols tls1.3
 }
 {{/if}}
 {{#if form.hsts}}


april:
This is actually correct; Caddy's documentation specifies that you need both a min and a max version.

I'll check with @mholt to see if the `max` is necessary, and if not, ask him to update the documentation to reflect that.  :)

angristan:
I just tried with `protocols tls1.3 tls1.3` and `protocols tls1.3` and both times TLS 1.2 was offered :thinking: I'm even more lost than before

mholt:
If only one value is specified, that becomes both the min and the max.

The v2 Caddyfile will use a single value as the min and leave the max as the default (which is currently TLS 1.3).

@angristan Feel free to open an issue at https://github.com/mholt/caddy if you're not seeing TLS 1.3 be used; be sure to post your full Caddyfile. Note that customizing cipher suites isn't allowed in TLS 1.3 so it might revert to TLS 1.2 if that's what you're doing.

angristan:
Nevermind! `protocols tls1.3 tls1.3` and `protocols tls1.3` both disable TLS 1.2. Sorry 🙂

april:
All that said, I still changed it to just be `protocols tls1.3`, because I had so many people ask this same question.  :)

PR #31, opened 2019-07-02
 Convert AWS ELB CloudFormation template from JSON to YAML 

### LINKED ISSUE #26: Change AWS CloudFormation to YAML

gene1wood:
Clean up AWS ELB

* Update policy name to include SSL recommendation version
* Improve parameter regex
* Improve documentation strings
* Output URL instead of DNS name
* Set syntax highlighting to yaml
* Set application version
  * The application version comes from the 2014-02-19 update to AWS ELB
    that adds support for TLS 1.2
    https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/DocumentHistory.html

Fixes #26


src/templates/partials/awselb.hbs
```diff
@@ -1,86 +1,51 @@
...
+      https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#genref-arns
+Resources:
+  ExampleELB:
+    Type: AWS::ElasticLoadBalancing::LoadBalancer
+    Properties:
+      Listeners:
+        - LoadBalancerPort: '443'
+          InstancePort: '80'
+          PolicyNames:
+            - Mozilla-{{form.config}}-v5-0
```
april: You can use `{{sstls.version}}` here, which will return 5, 5.1, 6, etc.

src/js/configs.js
```diff
@@ -14,8 +14,8 @@ module.exports = {
   // aws elb describe-load-balancer-policies --query "PolicyDescriptions[?PolicyName=='ELBSample-ELBDefaultCipherPolicy'].PolicyAttributeDescriptions[*].AttributeName[]"
   awselb: {
     hasVersions: false,
-    highlighter: 'json',
-    latestVersion: '2015.3',
+    highlighter: 'yaml',
+    latestVersion: '2014-02-19',
```
april: It's not going to break things here, but I try to keep all the versions as semantic versions, e.g. 2014.2.19.

src/templates/partials/awselb.hbs
```diff
@@ -1,86 +1,51 @@
...
+      Policies:
+        - PolicyName: Mozilla-{{form.config}}-v5-0
+          PolicyType: SSLNegotiationPolicyType
+          Attributes:
 {{#if (includes "SSLv3" output.protocols)}}
-                            {
-                                "Name": "Protocol-SSLv3",
-                                "Value": true
-                            },
+            - Name: Protocol-SSLv3
```
april: Looking back on this, it looks like it might be better to do an each loop over output.protocols, since their format matches what AWS uses.

src/templates/partials/awselb.hbs
```diff
@@ -1,86 +1,51 @@
...
-                                "Value": true
-                            },
-                            {
-                                "Name": "Server-Defined-Cipher-Order",
-                                "Value": true
-                            },
+            - Name: Protocol-TLSv1.2
+              Value: true
+            - Name: Server-Defined-Cipher-Order
+              Value: true
```
april: This actually needs to get fixed to no longer be true, whoops.

april: To clarify:

`{{#if output.serverPreferredOrder}}true{{else}}false{{/if}}`

src/templates/partials/awselb.hbs
gene1wood: Unfortunately in this specific case, AWS SSL policy names aren't allowed to have the `.` character so we could
* leave it hard coded
* produce an additional variable, in addition to `sstls.version` but which doesn't user periods
* do something here in the template to convert `.` to `-` (maybe we load a handlebar helper)?
* something I'm not thinking of.

What direction do you think we should go?

src/js/configs.js
gene1wood: Good idea, I'll change this.

src/templates/partials/awselb.hbs
gene1wood: Can I open an issue on that and solve it in a distinct PR?

src/templates/partials/awselb.hbs
gene1wood: We chatted and we're going to leave it like it is for now

src/templates/partials/awselb.hbs
april:
```
{{#each output.protocols}}
                            {
                                "Name": "{{this}}",
                                "Value": true
                            }
{{/each}}
```

gene1wood: I've made this change

src/templates/partials/awselb.hbs
gene1wood: I've made this change

april:
Looks good, thanks!

PR #32, opened 2019-07-02
Add support for AWS ALB

### LINKED ISSUE #25: Add AWS ALB HTTP to HTTPS redirect

gene1wood:
This adds support for the AWS Application Load Balancer

ALB doesn't support custom cipher lists
* "Application Load Balancers do not support custom security policies"
* https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html#describe-ssl-policies
As a result this maps our Old and Intermediate profiles to the closest similar
AWS policy.
* Old : ELBSecurityPolicy-TLS-1-0-2015-04
* Intermediate : ELBSecurityPolicy-TLS-1-2-2017-01
* Modern : unsupported

This CloudFormation template also contains a bunch of example infrastructure
which is needed to provision an ALB. I've added a comment to clarify that
for most people they'll be able to ignore all of it as they'll likely have
their own existing infrastructure.

The template includes an HTTP to HTTPS redirect, though HSTS isn't supported
at the ALB level (it could be done by the backend server though)

The server version comes from the 2018-07-25 revision to AWS ALB that added
support for redirects which we use to redirect HTTP to HTTPS
* https://docs.aws.amazon.com/elasticloadbalancing/latest/application/doc-history.html

Fixes #25


april:
Overall this looks good, but do you it would be good for it to perhaps change the "Supports" field underneath the title to say that it's merely the closest approximation as Mozilla (Whatever)?

gene1wood:
@april yes we should. I remember we'd talked and you said that the code supports overriding the client support string. I'll look for how to do that and add that to this PR

gene1wood:
I've resolved the conflicts with the added comments and HSTS conditional in 1c71c53ff45d1c8743def945606977acbea1d7a8

april:
Closing as this all got merged. Thanks, @gene1wood.  :)

PR #36, opened 2019-07-03
Add support for IIS on Windows 1809

lennybacon:
Here is support for IIS on Windows 10/2019 1809 and above


src/templates/partials/iis.hbs
```diff
@@ -0,0 +1,187 @@
...
+  $key = Get-Item -LiteralPath $path
+  if ($key.GetValue($name, $null) -ne $null) {
+    $r = Remove-ItemProperty -Path $path -Name $name
+  }
+}
+
+# Restrict the use of certain cryptographic algorithms and protocols in Schannel.dll
+# http://support2.microsoft.com/default.aspx?scid=kb;EN-US;245030
+Reset-Chiphers
+Set-Chipher -name "AES 128/128" -value 0xffffffff
```
april: Is this really how it is spelled?

src/templates/partials/iis.hbs
```diff
@@ -0,0 +1,187 @@
...
+function Remove-RegistryValue($path, $name){
+  Write-Host "Remove-RegistryValue ``$name``"
+  $key = Get-Item -LiteralPath $path
+  if ($key.GetValue($name, $null) -ne $null) {
+    $r = Remove-ItemProperty -Path $path -Name $name
+  }
+}
+
+# Restrict the use of certain cryptographic algorithms and protocols in Schannel.dll
+# http://support2.microsoft.com/default.aspx?scid=kb;EN-US;245030
```
april: This link isn't https, and it doesn't seem to work.

src/templates/partials/iis.hbs
```diff
@@ -0,0 +1,187 @@
...
+Set-Chipher -name "DES 56/56" -value 0x00000000
+Set-Chipher -name "NULL" -value 0x00000000
+Set-Chipher -name "RC2 128/128" -value 0x00000000
+Set-Chipher -name "RC2 40/128" -value 0x00000000
+Set-Chipher -name "RC2 56/128" -value 0x00000000
+Set-Chipher -name "RC4 128/128" -value 0x00000000
+Set-Chipher -name "RC4 40/128" -value 0x00000000
+Set-Chipher -name "RC4 56/128" -value 0x00000000
+Set-Chipher -name "RC4 64/128" -value 0x00000000
+Set-Chipher -name "Triple DES 168/168" -value 0x00000000
```
april: Is there no way to simply create a whitelist? This seems really inelegant.

src/js/configs.js
```diff
@@ -101,6 +101,54 @@ module.exports = {
...
+      'TLS_DHE_DSS_WITH_AES_256_CBC_SHA',
+      'TLS_DHE_DSS_WITH_AES_128_CBC_SHA',
+      'TLS_DHE_DSS_WITH_3DES_EDE_CBC_SHA',
+      'TLS_RSA_WITH_RC4_128_SHA',
+      'TLS_RSA_WITH_RC4_128_MD5',
+      'TLS_RSA_WITH_DES_CBC_SHA',
+      'TLS_DHE_DSS_WITH_DES_CBC_SHA',
+      'TLS_DHE_DSS_EXPORT1024_WITH_DES_CBC_SHA',
+      'TLS_RSA_WITH_NULL_MD5',
+      'TLS_RSA_EXPORT1024_WITH_RC4_56_SHA',
```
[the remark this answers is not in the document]
lennybacon: How else would you suggest to handle this?

src/js/configs.js
```diff
@@ -101,6 +101,54 @@ module.exports = {
...
+      'TLS_RSA_EXPORT_WITH_RC4_40_MD5',
+      'TLS_RSA_EXPORT1024_WITH_DES_CBC_SHA',        
+    ],
+    supportsHsts: true,
+    supportsOcspStapling: true,
+    tls13: noSupportedVersion,
```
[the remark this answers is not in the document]
lennybacon: No TLS 1.3 makes me sad too

src/js/index.js
```diff
@@ -11,11 +11,13 @@ import ini from 'highlight.js/lib/languages/ini';
 import json from 'highlight.js/lib/languages/json';
 import nginx from 'highlight.js/lib/languages/nginx';
 import yaml from 'highlight.js/lib/languages/yaml';
+import ps from 'highlight.js/lib/languages/powershell';
 hljs.registerLanguage('apache', apache);
 hljs.registerLanguage('ini', ini);
 hljs.registerLanguage('json', json);
 hljs.registerLanguage('nginx', nginx);
 hljs.registerLanguage('yaml', yaml);
+hljs.registerLanguage('ps', ps);
```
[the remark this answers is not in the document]
lennybacon: Sorted language registrations in a commit

src/templates/partials/iis.hbs
lennybacon: Fixed

src/templates/partials/iis.hbs
lennybacon: I could write a `Disable-Cipher`/`Enable-Cipher` method that takes an array `@("RC4 40/128", "RC4 56/128", ...)` as input...

april: I don't know, is there not a way to disable all ciphers and then enable them one by one? I'm not particularly familiar with IIS.

lennybacon: The values are keys in the registry. They are present and enabled by default :-(. I could write a method which iterates the keys and disables each without naming/listing each one. The a second method to enable the preferred ones. would this be fine?

src/js/configs.js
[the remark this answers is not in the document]
lennybacon: Would you prefer to drop all obvious insecure cipher suites (NULL, SHA1, CBC, etc...)?

src/templates/partials/iis.hbs
tomato42: that looks like a list of algorithms supported on the system level, not specifically for TLS...
are you sure that they need to be disabled for the TLS ciphers to be disabled?
I don't think we want to break backwards compatibility with PKCS#12 files if the user configures their stack TLS properly.

lennybacon: It is the configuration for [SChannel](https://docs.microsoft.com/en-us/windows/win32/secauthn/secure-channel) which is used for encrypted communication. 

> Microsoft TLS/SSL Security Provider, the Schannel.dll file, uses the CSPs that are listed here to conduct secure communications over SSL or TLS in its support for Internet Explorer and Internet Information Services (IIS). 
[Source](https://support.microsoft.com/en-us/help/245030/how-to-restrict-the-use-of-certain-cryptographic-algorithms-and-protoc)

This will also affect apps running on the platform - e.g. PowerShell, Outlook and so on **when communication encrypted over the wire**.

What do you mean by *backwards compatibility with PKCS#12 files*?

tomato42: > What do you mean by backwards compatibility with PKCS#12 files?

I was thinking of ciphers available in TLS 1.0, where a 128 bit RC2 was never defined, but I forgot about SSLv2 ciphers. At the same time, there are RC2 ciphers defined for use in PKCS#12 files and they have the same key sizes as SSLv2 ciphers.

lennybacon: SChannel has nothing to do with the crypto operations on files.

src/js/configs.js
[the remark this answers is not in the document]
lennybacon: @april what's to do for me here?

lennybacon:
Dear @april is the generator dead? The PR and my question what needs to be done from your perspective were unanswered for month...

april:
No, it's not dead, I just don't know if I feel comfortable recommending a giant blob of command line commands to run, especially without a good way to undo it.

NoNameForMee:
I have not looked through the full details of this one MR (which appears related to issue #54), however I came across something similar by Alexander Hass, available at: https://www.hass.de/content/setup-microsoft-windows-or-iis-ssl-perfect-forward-secrecy-and-tls-12

The script provided over there includes support for other installed OS & IIS versions as well as a sort of "reset" script. Perhaps it would be possible to make some similar here as well... Or perhaps reach out to Alexander Hess and have him help out/contribute (as the script on his website is strictly speaking copyrighted by Alexander Hess).

lennybacon:
Would be great to keep this going. If you like further assistance we could also reach out to MSFT folks...

april:
If Microsoft has a recommended and supported way to do (and undo) all this, I'd happily add it.

I just don't want people to yell at me if a bunch of scripts screw up their machine.  :)

arnydo:
> 
> 
> If Microsoft has a recommended and supported way to do (and undo) all this, I'd happily add it.
> 
> I just don't want people to yell at me if a bunch of scripts screw up their machine. :)

Would it be more appropriate to simply provide the list of recommended ciphers so that they can be input into the appropriate Group Policy? Here is the documentation on setting the [supported ciphers via GPO](https://docs.microsoft.com/en-us/windows-server/security/tls/manage-tls).

The generator could return the properly formatted, comma-separated, list of ciphers that can be copied and pasted into the GPO. 

Example:
Generator returns:

```
TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_DHE_RSA_WITH_AES_256_GCM_SHA384,TLS_DHE_RSA_WITH_AES_128_GCM_SHA256
```
Paste into GPO:
image

lennybacon:
I contacted some ppl at MSFT but there seems to be nobody feeling responsible as there is a overlap of IIS and SCHANNEL. Also nobody I contacted saw the need/goal/benefit in this project so we cannot expect support from them.

I really like the idea of GPOs, as it is the default management/administration path and is also enterprise compatible.

With the GPOs its is also possible to [create an XML file](https://social.technet.microsoft.com/Forums/windows/en-US/ac40c188-1e98-45a7-a7bb-03092906aadf/exportimport-of-group-policy-preferences-via-xml)

arnydo:
I agree, I feel this is an excellent resource to be able to go to and retrieve the latest "recommended" TLS settings for each of the service that we may have deployed. Considering the uniqueness of IIS/SCHANNEL/Windows there should really be, at a minimum, a listing of the current recommended settings. This doesn't necessarily have to be the actual XML file of the settings or a series of PowerShell scripts, but I think it should certainly be considered.

I do like the XML idea. Could simply have the template spit out the XML data that can be copied into the Registry section of the GPO.

april:
I mean, the recommendations are both in the JSON file that can be programatically interacted with as well as the Server Side TLS Guidelines:

https://wiki.mozilla.org/Security/Server_Side_TLS

What would the listing say that isn't already in the article?

arnydo:
> 
> 
> I mean, the recommendations are both in the JSON file that can be programatically interacted with as well as the Server Side TLS Guidelines:
> 
> https://wiki.mozilla.org/Security/Server_Side_TLS
> 
> What would the listing say that isn't already in the article?

Honestly, I didn't notice the JSON version. That would suffice in our case. Considering there doesn't seem to be a `simple` method to address IIS/SCHANNEL/Windows in a clean way.

Thanks for pointing this out @april .

nemchik:
This kind of feels like it might be more successful as a PowerShell module that could be installed from the PowerShell gallery. You may also have a look at https://www.nartac.com/Products/IISCrypto/

Firefishy:
I agree with @nemchik , the https://www.nartac.com/Products/IISCrypto/ is a much better way to manage this on Windows. Giving users a large blob of powershell is just too scary with no safe revert if something goes wrong.

My humble suggestion would be to list windows as an option, but add some text similar to https://github.com/mozilla/ssl-config-generator/blob/master/src/templates/partials/nosupport.hbs and suggest they look at https://www.nartac.com/Products/IISCrypto/

nemchik:
> I agree with @nemchik , the https://www.nartac.com/Products/IISCrypto/ is a much better way to manage this on Windows. Giving users a large blob of powershell is just too scary with no simple/safe revert if something goes wrong.
> 
> My humble suggestion would be to list windows as an option, but add some text similar to https://github.com/mozilla/ssl-config-generator/blob/master/src/templates/partials/nosupport.hbs and suggest they look at https://www.nartac.com/Products/IISCrypto/

Just to add a bit, I love IISCrypto, but I don't think it should be relied on as a mechanism to apply settings from here. Sure it probably could work, but since there's no telling how that program might change over the years and how that might affect its ability to apply settings provided here I would not view it as a viable long term solution. Powershell on the other hand should be pretty straightforward. Having the option to apply changes or revert to Windows built in standard settings should be plenty.

Running a powershell script is no more complicated than configuring apache or nginx.

Firefishy:
> Just to add a bit, I love IISCrypto, but I don't think it should be relied on as a mechanism to apply settings from here. Sure it probably could work, but since there's no telling how that program might change over the years and how that might affect its ability to apply settings provided here I would not view it as a viable long term solution. Powershell on the other hand should be pretty straightforward. Having the option to apply changes or revert to Windows built in standard settings should be plenty.
> 
> Running a powershell script is no more complicated than configuring apache or nginx.

My suggestion would be to add an `IIS` radio button and have something like the following text:
```
# IIS server SSL configuration cannot be set using a text configuration file and is unsupported by this tool.
#
# But there are other options available:
# 
#   IISCrypto: https://www.nartac.com/Products/IISCrypto/ (easiest)
#   Powershell Script: https://www.hass.de/content/setup-microsoft-windows-or-iis-ssl-perfect-forward-secrecy-and-tls-12
#   Microsoft Documentation: https://docs.microsoft.com/en-us/troubleshoot/windows-server/windows-security/restrict-cryptographic-algorithms-protocols-schannel 
```

lennybacon:
Beside the fact that IISCrypto is a handy tool...

IMHO a closed source tool is always more **scary** that scripts or open source where I can see what it actually does.

🤷‍♂️

lennybacon:
I'll keep an eye on this [issue](https://github.com/dotnet/aspnetcore/issues/9349) as it is probably related.

Beside that I totally fine to close this thread.

PR #39, opened 2019-07-10
use correct array in MySQL template

Manawyrm:
with this PR, the ssl-cipher= line isn't displayed when the ciphers array is empty.


april:
This looks great, but could you please remove the extra newline at the end of the file? Thanks!

Manawyrm:
Yes, I hope it's okay now.

april:
Thanks!

PR #41, opened 2019-07-12
fix postfix template - ciphers vs ciphersuites

scintill:
I haven't tested this, but I think both should say ciphers. It currently generates "tls_medium_cipherlist=" (empty) which postfix doesn't accept.

Also, I don't see the postfix source calling `SSL*_set_ciphersuites`, so I guess they don't support specifying 1.3 ciphersuites (only ciphers of TLS1.2 and below.)


april:
This looks great, but could you please remove the change to the newline at the end of the file?

Thanks!

scintill:
OK, I took out the newline.

april:
Thanks!

PR #81, opened 2020-02-13
Update Intermediate security policy for AWS ALB

sonicdoe:
Updates the Intermediate configuration for Amazon Web Services’s Application Load Balancer to `ELBSecurityPolicy-FS-1-2-Res-2019-08` [which was added on October 8, 2019](https://aws.amazon.com/about-aws/whats-new/2019/10/application-load-balancer-and-network-load-balancer-add-new-security-policies-for-forward-secrecy-with-more-strigent-protocols-and-ciphers/).

This policy is the same as the previous `ELBSecurityPolicy-TLS-1-2-2017-01` but removes support for AES128-GCM-SHA256, AES128-SHA256, AES256-GCM-SHA384, and AES256-SHA256.

Here’s a table comparing the TLS 1.2 cipher suites of the [Intermediate configuration](https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28recommended.29), the previous `ELBSecurityPolicy-TLS-1-2-2017-01` security policy, and the new `ELBSecurityPolicy-FS-1-2-Res-2019-08` security policy:

| Intermediate                  | TLS 1.2                       | FS 1.2 Res                    |
|-------------------------------|-------------------------------|-------------------------------|
| ECDHE-ECDSA-AES128-GCM-SHA256 | ECDHE-ECDSA-AES128-GCM-SHA256 | ECDHE-ECDSA-AES128-GCM-SHA256 |
| ECDHE-RSA-AES128-GCM-SHA256   | ECDHE-RSA-AES128-GCM-SHA256   | ECDHE-RSA-AES128-GCM-SHA256   |
| ECDHE-ECDSA-AES256-GCM-SHA384 | ECDHE-ECDSA-AES128-SHA256     | ECDHE-ECDSA-AES128-SHA256     |
| ECDHE-RSA-AES256-GCM-SHA384   | ECDHE-RSA-AES128-SHA256       | ECDHE-RSA-AES128-SHA256       |
| ECDHE-ECDSA-CHACHA20-POLY1305 | ECDHE-ECDSA-AES256-GCM-SHA384 | ECDHE-ECDSA-AES256-GCM-SHA384 |
| ECDHE-RSA-CHACHA20-POLY1305   | ECDHE-RSA-AES256-GCM-SHA384   | ECDHE-RSA-AES256-GCM-SHA384   |
| DHE-RSA-AES128-GCM-SHA256     | ECDHE-ECDSA-AES256-SHA384     | ECDHE-ECDSA-AES256-SHA384     |
| DHE-RSA-AES256-GCM-SHA384     | ECDHE-RSA-AES256-SHA384       | ECDHE-RSA-AES256-SHA384       |
|                               | AES128-GCM-SHA256             |                               |
|                               | AES128-SHA256                 |                               |
|                               | AES256-GCM-SHA384             |                               |
|                               | AES256-SHA256                 |                               |

### DIFF
--- a/src/templates/partials/awsalb.hbs
+++ b/src/templates/partials/awsalb.hbs
@@ -41,7 +41,7 @@ Resources:
       LoadBalancerArn: !Ref ExampleALB
       Port: 443
       Protocol: HTTPS
-      SslPolicy: {{#if (includes "TLSv1" output.protocols)}}ELBSecurityPolicy-TLS-1-0-2015-04{{else}}ELBSecurityPolicy-TLS-1-2-2017-01{{/if}}
+      SslPolicy: {{#if (includes "TLSv1" output.protocols)}}ELBSecurityPolicy-TLS-1-0-2015-04{{else}}ELBSecurityPolicy-FS-1-2-Res-2019-08{{/if}}
 {{#if form.hsts}}
 
   # {{form.serverName}} doesn't support HSTS, but it can redirect to HTTPS


april:
Gene, this looks fine to me, but I thought I'd run it by you to see if you had any caveats.

jrchamp:
I have very little love for old versions of Windows or Internet Explorer, but I do want to note that's why the DHE-RSA suites are present in Intermediate. If you're okay dropping them, then at least you're doing it on purpose and not on accident. :heart:

sonicdoe:
I fully agree with keeping compatibility with Internet Explorer 11 on Windows 7 as specified for the Intermediate configuration. Looking at https://github.com/mozilla/server-side-tls/issues/178#issuecomment-506146146, another requirement seems to be keeping compatibility even if the server uses a certificate with an RSA key.

For AWS ALB, I think this is taken care of by ECDHE-RSA-AES128-SHA256 which is not present in the Intermediate configuration but is present in both the old and new security policy. The corresponding IANA name, TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256, is listed on [Qualys SSL Labs’s user agent capabilities page](https://www.ssllabs.com/ssltest/viewClient.html?name=IE&version=11&platform=Win%207&key=143).

april:
Yep, that was my thought as well. Basically they prefer `ECDHE-RSA-AES128-SHA256` over `DHE-RSA-AES128-GCM-SHA256`, for IE11 support. I don't think either one of them is particularly wrong: they're both somewhat outdated cipher suites and you have to pick one to keep around.

jrchamp:
Sorry! I missed the fact that the table is not aligned. I read the first column (which is 100% FS and AEAD) and didn't realize the second and third columns were completely different.

**This change is a definite improvement** because nothing at the Intermediate level is being lost by this change (comparing the before "middle" and the after "right" columns). Not that they're likely to make changes after the fact, but it would make me a bit happier if @awslabs dropped ECDHE-ECDSA-AES128-SHA256 and ECDHE-ECDSA-AES256-SHA384 from that preset.

sonicdoe:
Sorry, I definitely see how the table can be confusing. To be clear, the old and new security policies are the same except for the new one no longer including AES128-GCM-SHA256, AES128-SHA256, AES256-GCM-SHA384, and AES256-SHA256 (the last four in the middle column).

As to the Intermediate configuration, the security policy shares the first four cipher suites but does not include any ChaCha20-Poly1305 and DHE cipher suites (the last four in the first column). Instead, it includes ECDHE-ECDSA-AES128-SHA256, ECDHE-RSA-AES128-SHA256, ECDHE-ECDSA-AES256-SHA384, and ECDHE-RSA-AES256-SHA384.

[Looking at the AWS documentation](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html#describe-ssl-policies), there’s no security policy stricter than this.

april:
I haven't heard from @gene1wood, but I think my concerns have been allayed here.  :)

PR #82, opened 2020-02-13
Fix link to AWS ALB security policies

sonicdoe:
The previous URL linked to security policies for Classic Load Balancers, however, this file is about Application Load Balancers.


april:
Thanks!

Issue #86, opened 2020-03-11
Include Redis open-source in configuration generator


IAmATeaPot418:
Redis open-source 6 will include many TLS enhancements in its upcoming release.

I'm happy to submit a PR with the guidance but is there any process or guidelines to help accomplish this or a review process?

I'm looking mostly for acceptance criteria so I can bake it into the PR.

april:
You can learn the process to add new software here:

https://github.com/mozilla/ssl-config-generator#adding-new-software

IAmATeaPot418:
@april I've submitted https://github.com/mozilla/ssl-config-generator/pull/87 to add support for Redis.

LMK what you think. Please don't merge this yet. This version is not GA and the release candidate is still being tested.

PR #87, opened 2020-03-12
Adding Redis 6.0 Support

IAmATeaPot418:
Adding Redis 6.0 support. Please do not merge yet. This version is in release candidate and is not officially GA. If there are any material changes I will update this PR. I'll let you know when released.

### DIFF
--- a/src/templates/partials/redis.hbs
+++ b/src/templates/partials/redis.hbs
@@ -1,0 +1,31 @@
+# {{output.header}}
+# {{{output.link}}}
+
+# Running Redis with TLS requires you to build Redis specifically for TLS
+# This can be done when making redis using "make && make BUILD_TLS=yes"
+
+################################# TLS/SSL #####################################
+port 0
+tls-port 6379
+tls-cert-file redis.crt
+tls-key-file redis.key
+tls-ca-cert-file ca.crt
+tls-ca-cert-dir /etc/ssl/certs
+tls-auth-clients yes
+tls-replication yes
+tls-cluster yes
+{{#unless (includes "old" form.config)}}
+# Running Redis with TLS 1.3 requires OpenSSL 1.1.1 or greater
+{{/unless}}
+{{#if (includes "intermediate" form.config)}}
+tls-protocols "TLSv1.2 TLSv1.3"
+{{/if}}
+{{#if (includes "modern" form.config)}}
+tls-protocols TLSv1.3
+{{/if}}
+{{#if (includes "intermediate" form.config)}}
+tls-ciphers ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384
+{{/if}}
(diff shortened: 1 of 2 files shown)


IAmATeaPot418:
This PR closes https://github.com/mozilla/ssl-config-generator/issues/86

src/templates/partials/redis.hbs
```diff
@@ -0,0 +1,31 @@
...
+port 0
+tls-port 6379
+tls-cert-file redis.crt
+tls-key-file redis.key
+tls-ca-cert-file ca.crt
+tls-ca-cert-dir /etc/ssl/certs
+tls-auth-clients yes
+tls-replication yes
+tls-cluster yes
+{{#unless (includes "old" form.config)}}
```
april: The site should automatically warn users if this is the case, so this shouldn't be necessary.

src/templates/partials/redis.hbs
```diff
@@ -0,0 +1,31 @@
...
+# Running Redis with TLS 1.3 requires OpenSSL 1.1.1 or greater
+{{/unless}}
+{{#if (includes "intermediate" form.config)}}
+tls-protocols "TLSv1.2 TLSv1.3"
```
april: All of these should be programatically generated, there should in general not be any manual setting based on things. This is so all the profiles update once the JSON file is updated.

src/templates/partials/redis.hbs
```diff
@@ -0,0 +1,31 @@
...
+{{/if}}
+{{#if (includes "modern" form.config)}}
+tls-protocols TLSv1.3
+{{/if}}
+{{#if (includes "intermediate" form.config)}}
+tls-ciphers ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384
+{{/if}}
+{{#if (includes "modern" form.config)}}
+tls-prefer-server-ciphers yes
```
april: Same here, etc. Look at the other profiles to see how this is done.

IAmATeaPot418: @april just to confirm on this one and the same for the other two comments you made, the ask is to use {{{join output.ciphers ":"}}} (for example) rather than hardcoding the cipher suites and ciphers correct?

I actually happened to pick the only one that hardcoded them (caddy.hbs) as my reference point. Might actually just update that in my PR too at this point so nobody makes my mistake (assuming this is correct).

src/templates/partials/redis.hbs
IAmATeaPot418: output.protocols instead should be used. Got it.

IAmATeaPot418:
@april - My most recent commit should have added a completely dynamic generation based on the pre-defined fields.

LMK what you think.

IAmATeaPot418:
Ping @april

april:
Looks good, I'll probably make some tweaks but overall very nice work.  :)

IAmATeaPot418:
Thanks, April! It currently looks like this will be GA around the end of the month.

april:
Okay, everything should be live now. I made a lot of tweaks, but I really appreciate you putting in the hard work on this.  :)

IAmATeaPot418:
hey @april - If the TLS protocols use more than one they must be surrounded in quotations to take. Could we add the quotations back in?

PR #88, opened 2020-03-13
Updating PostgreSQL template for version 12

crispygoth:
PostgreSQL version 12 added a new ssl_min_protocol_version configuration
option to control the SSL versions the server uses.

When testing I also discovered that PostgreSQL rejects an empty
ssl_ciphers value, so I've wrapped this in an if to prevent it being
output if ciphers is empty (modern profile).


april:
Looks good, thank you!

april:
This should be live now. Thanks again!

PR #90, opened 2020-03-24
simplify lighttpd config

gstrauss:
simplify lighttpd config

* latest version is lighttpd 1.4.55  (how does docs/39dd7eed14622039aa44.index.js get updated?)
* lighttpd 1.4.50 and later has simpler (and more efficient) syntax for http to https redirect
* `protocol = "https://"` is not use by lighttpd.conf and gets ignored
* `ssl.disable-client-renegotiation = "enable"` is the default, and always has been for that config option.
* `setenv.add-environment = ( "HTTPS" => "on" )` is unnecessary, as lighttpd already does this for backends when the connection uses https.

Greetings from a lighttpd developer. :)

### DIFF
--- a/src/js/configs.js
+++ b/src/js/configs.js
@@ -78,7 +78,7 @@ module.exports = {
   },
   lighttpd: {
     highlighter: 'nginx',
-    latestVersion: '1.4.54',
+    latestVersion: '1.4.55',
     name: 'lighttpd',
     tls13: '1.4.53'
   },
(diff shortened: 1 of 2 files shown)


april:
Looks good, thanks! docs gets updated when you do a build with npm, so no need to worry about that.

april:
This should be live now, I also cleaned up the lighttpd configuration a bit.

I appreciate you submitting here, it would be great if every one of the supported pieces of software did.  :)

PR #91, opened 2020-04-04
The quotes are important :)

IAmATeaPot418:
The quotation marks around TLS protocols prevent Redis from interpreting each individual protocol as a separate argument and groups all of them together.


april:
Could you contact the redis people and have them update the documentation? The `redis.conf` example is super unclear.

IAmATeaPot418:
Sure thing @april  I'll submit a PR to fix.

PR #94, opened 2020-04-27
chore: Remove obsolete postfix parameter `smtpd_use_tls`

polarathene:
`smtpd_tls_security_level` has replaced this config parameter. `smtpd_use_tls` has no relevance in the config anymore.

See: http://www.postfix.org/postconf.5.html#smtpd_tls_security_level

> this overrides the obsolete parameters `smtpd_use_tls` and `smtpd_enforce_tls`.


april:
Could you fix up the newline change at the end of the file please?

april:
I don't think your commit made any changes? 😅

polarathene:
[Github doesn't seem to support doing so and is enforcing it](https://medium.com/@gogasca_/github-editor-and-not-new-line-in-jupyter-notebooks-2cbece23839b) :disappointed:  I'll need to edit it locally I guess.

Any reason for this requirement btw? It's not mentioned on the README and I think I recall you asking someone else before to do the same?

[Reasons for keeping it](https://unix.stackexchange.com/a/18789):

> A text file, under unix, consists of a series of lines, each of which ends with a newline character (\n). A file that is not empty and does not end with a newline is therefore not a text file.

> POSIX, this is a set of standards specified by IEEE to maintain compatibility between operating systems.
> One of which is the definition of a "line" being a sequence of zero or more non- characters plus a terminating newline character.
>So for that last line to be recognised as an actual "line" it should have a terminating new line character.

april:
The templates render that extra line if they're there, so it needs to not be inside any of them.  :)

polarathene:
Resolved.

PR #101, opened 2020-06-20
Update Go generator

FiloSottile:


FiloSottile:
The Old config also prioritizes some cipher suites that we disable by default because they don't implement any Lucky13 countermeasures.

https://pkg.go.dev/crypto/tls?tab=doc#pkg-note-BUG

 AFAIK they don't really move the needle in terms of compatibility so you could remove them, but at least they should be sorted last. Is there interest in making a custom order for Go?

april:
@FiloSottile, thanks so much!

PR #105, opened 2020-07-16
Use noSupportedVersion since dovecot doesn't support TLSv1.3 for now

fabiopozzi:
Since the fix to allow Dovecot to recognize the option to make use of TLSv1.3 it has not been merged for now, I think we should disable the "modern" configuration since it leads to a non-working config file.


april:
Thank you!

PR #119, opened 2020-10-07
Adding stunnel support

wevsty:
Stunnel can be used to provide secure encrypted connections for clients or servers that do not speak TLS natively.
So I added support for stunnel.

PR tesed on ubuntu 20.04


src/templates/partials/stunnel.hbs
```diff
@@ -0,0 +1,35 @@
+# {{output.header}}
+# {{{output.link}}}
+{{#if (minver "5.40" form.serverVersion)}}
```
gene1wood, line 3: It looks like the `if` and `/if` clauses don't match as if you set the version below 5.40 this produces an empty config.

gene1wood: Also, is `5.40` the earliest support for TLS1.3? The release notes for [Version 5.45, 2018.05.21, urgency: MEDIUM](https://www.stunnel.org/NEWS.html) make it look like TLS1.3 shows up in `5.45`. Can you share where the `5.40` number comes from?

src/templates/partials/stunnel.hbs
```diff
@@ -0,0 +1,35 @@
...
+{{#if (minver "1.1.1" form.opensslVersion)}}
+sslVersionMin = TLSv1.3
+ciphersuites = {{{join output.cipherSuites ":"}}}
+{{else}}
+  {{#if (minver "1.1.0" form.opensslVersion)}}
+sslVersionMin = TLSv1.2
```
gene1wood, line 9: It looks like in [Version 4.54, 2012.10.09, urgency: MEDIUM](https://www.stunnel.org/NEWS.html) TLSv1.2 is supported with OpenSSL 1.0.1
> New parameters to configure TLS v1.1/v1.2 with OpenSSL version 1.0.1 or higher (thx to Henrik Riomar).

Can you share where the mapping of TLSv1.2 and openssl 1.1.0 comes from?

gene1wood (changes requested):

src/js/configs.js
```diff
@@ -157,6 +157,16 @@ module.exports = {
...
     usesOpenssl: true,
   },
+  stunnel: {
+    highlighter: 'ini',
+    latestVersion: '5.56',
+    name: 'stunnel',
+    showSupports: false,
+    supportsHsts: false,
+    supportsOcspStapling: false,
+    tls13: '5.56',
```
gene1wood, line 167: It sounds like this is either `5.40` like you use in the template or `5.45` as it appears to be in the release notes.

src/templates/partials/stunnel.hbs
wevsty: My intention was not to support a version that was too old.
Considering that Ubuntu 16.04 LTS will reach its lifecycle in April 2021, and the version available in the Ubuntu 16.04 repository is ```5.39```,So i chose not to export anything below ```5.40```.

But since the question was raised, I'll modify the logic, which will allow to select the lower version.

src/templates/partials/stunnel.hbs
wevsty: According to the documentation of the configuration file
sslVersionMin This option is only available when compiling with OpenSSL 1.1.0 or higher.
If compiling with a version of openssl lower than 1.1.0 should use an option like "options = NO_TLSv1".

Link:
https://www.stunnel.org/static/stunnel.html

wevsty:
@gene1wood 

I have modified the code and will now support versions prior to ```5.40```.
I also reconfirmed the sslVersionMin option, which was added from version ```5.50``` onwards.
Regarding TLS 1.3 support, OpenSSL is supported from version ```1.1.1```, And stunnel ```5.50``` starts using ```OpenSSL 1.1.1``` by default so set the minimum version of TLS 1.3 to  stunnel ```5.50```

```
Version 5.50, 2018.12.02, urgency: MEDIUM

    New features
        32-bit Windows builds replaced with 64-bit builds.
        OpenSSL DLLs updated to version 1.1.1.
        Check whether "output" is not a relative file name.
        Added sslVersion, sslVersionMin and sslVersionMax for OpenSSL 1.1.0 and later.
```

jax7778:
Sorry to necro this, but I was just wondering, are there any plans to support stunnel as a config? It looked like this was pretty close?

wevsty:
@jax7778 
I updated the version number of the latest version. 
And I re-tested it and I think this PR still works and merges properly.

If there are any other questions you can contact me to modify the PR

jax7778:
@wevsty 

Thanks so much for doing this, I know those of that us that make use of stunnel will be grateful.

jax7778:
Sorry to be a bother, but is this planned on being merged into the main Mozilla SSL Config Page? I just would like to confirm, thanks for all you are doing here.

src/templates/partials/stunnel.hbs
gene1wood: Got it, ya the docs say

> Use sslVersionMax or sslVersionMin option instead of disabling specific TLS protocol versions when compiled with OpenSSL 1.1.0 or later.

gene1wood (approved):

gene1wood:
Thanks @wevsty for the PR and thanks @jax7778 and @wevsty for your patience. This is deployed

janbrasna (commented):
I'm not sure the outputs reflect chosen configuration. For the default render (`intermediate`, `latestVersion`) I seem to be getting `modern` no matter what config I select — unless I manually downgrade the server version below `tls13` when TLSv1.2-related output starts appearing reflecting parts of the chosen configuration level.

@gene1wood Is that as intended? It seems to hard-code the most restrictive flags based on the serverVersion and opensslVersion available, not taking the flexibility of the configs into account. If you confirm this is indeed not as it should be please let me know, I'll open an issue and maybe a draft PR to start adapting it to what it probably should have been(?)

gene1wood:
@janbrasna Would you open an issue with these suggested changes so they don't get lost here in this closed PR?

jax7778:
> Thanks @wevsty for the PR and thanks @jax7778 and @wevsty for your patience. This is deployed

Thank you for all the work!

src/templates/partials/stunnel.hbs
```diff
@@ -0,0 +1,45 @@
...
+    {{#if (minver "5.50" form.serverVersion)}}
+        {{#if (minver "1.1.0" form.opensslVersion)}}
+            {{#if (minver "1.1.1" form.opensslVersion)}}
+sslVersionMin = TLSv1.3
+ciphersuites = {{{join output.cipherSuites ":"}}}
+            {{ else }}
+sslVersionMin = TLSv1.2
+ciphers = {{{join output.ciphers ":"}}}
+            {{/if}}
+        {{ else }}
```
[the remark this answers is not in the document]
janbrasna: - testing for `opensslVersion` for TLSv1.3 is not to be done in templates, the app itself does that and populates the `output` state accordingly
- cipher suites for TLSv1.3 are to be left to OpenSSL/system defaults
- the `output.*` array join lines only to be rendered when there are any items in them not to produce empty cipher strings (=disabling all)
- most of the other conditionals should test for presence in `output` that's been prepared by the app, not in `form` versions entered by the user (only version conditionals should be for disabling lines the server won't understand; I'll have to xref this a bit more…)
- there's a mix of sslVersion, sslVersionMin and options that all achieve the same; I'd stick with other templates' pattern to iterate over `NO_*` ops unless they're in the `output` and leave out the other lines

(+ I'm tempted to add some recommended config lines to chroot and run under nobody, +skipping some TCP algos as in [RHEL example](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/security_guide/sec-configuring_stunnel_wrapper#doc-wrapper).)

PR #120, opened 2020-10-13
lighttpd updates

gstrauss:
refresh lighttpd config

* Comment how to listen on multiple sockets, IPv4 and IPv6.
* Disable Session Tickets.
* Add warnings for using old versions of OpenSSL.
* Use separate blocks to redirect from http to https, and to set Strict-Transport-Security response header. (reduce duplication of config directives)

These changes prepare for the upcoming release of lighttpd 1.4.56, which will be submitted in a separate pull request, once lighttpd 1.4.56 is released in the next month or so.

### DIFF
--- a/src/templates/partials/lighttpd.hbs
+++ b/src/templates/partials/lighttpd.hbs
@@ -1,7 +1,13 @@
 # {{output.header}}
 # {{{output.link}}}
+#server.modules += ("mod_redirect")
+#server.modules += ("mod_setenv")
+#server.modules += ("mod_openssl")
+#server.port = 80
+$SERVER["socket"] == "[::]:80" { }
+
 {{#if form.hsts}}
-$SERVER["socket"] == ":80" {
+$HTTP["scheme"] == "http" {
 {{#if (minver "1.4.50" form.serverVersion)}}
     url.redirect = ("" => "https://${url.authority}${url.path}${qsa}")
 {{else}}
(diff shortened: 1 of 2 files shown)


src/templates/partials/lighttpd.hbs
```diff
@@ -1,7 +1,11 @@
 # {{output.header}}
 # {{{output.link}}}
+$SERVER["socket"] ==     ":80" { }
```
april, line 3: This only does IPv4?

src/templates/partials/lighttpd.hbs
```diff
@@ -1,7 +1,11 @@
...
+$SERVER["socket"] == "[::]:80" { }
+
 {{#if form.hsts}}
-$SERVER["socket"] == ":80" {
+#server.modules += ("mod_redirect")
```
april, line 7: If we require a module that's not included by default, mention it at the top.

src/templates/partials/lighttpd.hbs
```diff
@@ -11,36 +15,59 @@ $SERVER["socket"] == ":80" {
...
+}
 {{/if}}
+
+#server.modules += ("mod_openssl")
 $SERVER["socket"] == ":443" {
-    ssl.engine   = "enable"
+    ssl.engine  = "enable"
 
+  {{#if (minver "1.4.53" form.serverVersion)}}
+    ssl.privkey = "/path/to/private_key"
```
april, line 32: Do the equal signs still line up after all this?

src/templates/partials/lighttpd.hbs
```diff
@@ -11,36 +15,59 @@ $SERVER["socket"] == ":80" {
...
+    ssl.pemfile = "/path/to/signed_cert"
+    ssl.ca-file = "/path/to/intermediate_certificate"
+    # alternative: pemfile is cert+privkey, ca-file is the intermediate chain in one file
+    #ssl.pemfile = "/path/to/signed_cert_plus_private_key"
+    #ssl.ca-file = "/path/to/intermediate_certificate"
```
april, line 37: We try to be opinionated, just pick one option and present it to them.

src/templates/partials/lighttpd.hbs
```diff
@@ -11,36 +15,59 @@ $SERVER["socket"] == ":80" {
...
     {{/if}}
 {{/if}}
 
     # {{form.config}} configuration
-{{#if (minver "1.4.48" form.serverVersion)}}
-    ssl.openssl.ssl-conf-cmd = ("Protocol" => "ALL, -SSLv2, -SSLv3{{#unless (includes "TLSv1" output.protocols)}}, -TLSv1{{/unless}}{{#unless (includes "TLSv1.1" output.protocols)}}, -TLSv1.1{{/unless}}{{#unless (includes "TLSv1.2" output.protocols)}}, -TLSv1.2{{/unless}}")
-{{else}}
-    # Please upgrade to 1.4.48 or else you cannot fully disable deprecated protocols
+ {{#if (minver "1.1.0" form.opensslVersion)}}
+  {{#if (minver "1.4.55" form.serverVersion)}}
```
april, line 53: Why does this syntax not work with older versions?

src/templates/partials/lighttpd.hbs
```diff
@@ -11,36 +15,59 @@ $SERVER["socket"] == ":80" {
...
-    setenv.add-response-header  = (
-        "Strict-Transport-Security" => "max-age={{output.hstsMaxAge}}"
-    )
-{{/if}}
-}
\ No newline at end of file
+  {{/if}}
+ {{else}}
+    # Please upgrade to openssl 1.1.1 or later
+    # Earlier openssl versions no longer receive security patches from openssl.org
```
april, line 64: Please remove comments like this, as it's not really in the purview of this tool.  :)

src/templates/partials/lighttpd.hbs
```diff
@@ -11,36 +15,59 @@ $SERVER["socket"] == ":80" {
...
+    ssl.use-sslv2 = "disable"
+    ssl.use-sslv3 = "disable"
+ {{/if}}
+    ssl.cipher-list = "{{{join output.ciphers ":"}}}"
+    # ssl.honor-cipher-order = "disable" respects client cipher preference
+    # (safe when ssl.cipher-list is restricted to only ciphers which support Perfect Forward Secrecy)
+    ssl.honor-cipher-order = "{{#if output.serverPreferredOrder}}enable{{else}}disable{{/if}}"
+}
+#$SERVER["socket"] == "[::]:443" { ... } # repeat entire $SERVER["socket"] == ":443" { ... } config above for IPv6
```
april, line 73: Why is it not possible to have a config that works for both IPv4 and IPv6? I find it hard to believe it requires you to repeat all of your config twice.

src/templates/partials/lighttpd.hbs
```diff
@@ -11,36 +15,59 @@ $SERVER["socket"] == ":80" {
...
\ No newline at end of file
+  {{/if}}
+ {{else}}
+    # Please upgrade to openssl 1.1.1 or later
+    # Earlier openssl versions no longer receive security patches from openssl.org
+    ssl.use-sslv2 = "disable"
+    ssl.use-sslv3 = "disable"
+ {{/if}}
+    ssl.cipher-list = "{{{join output.ciphers ":"}}}"
+    # ssl.honor-cipher-order = "disable" respects client cipher preference
```
april, line 69: This is for the guidelines, not for the generator.  :)

src/templates/partials/lighttpd.hbs
gstrauss: Yes, it only does IPv4.
Also, since the default is `server.port = 80`, I'm going to push a change to comment this line out.

src/templates/partials/lighttpd.hbs
gstrauss: ok.  will move those comments to the top

src/templates/partials/lighttpd.hbs
gstrauss: many of them do.

src/templates/partials/lighttpd.hbs
gstrauss: ok, updated

src/templates/partials/lighttpd.hbs
gstrauss: openssl 1.0.2 added API `SSL_CONF_cmd` including "Protocol"
lighttpd 1.4.48 added `ssl.openssl.ssl-conf-cmd`
openssl 1.1.0 added API `SSL_CONF_cmd` "MinProtocol"

I have adjusted the conditions to better reflect those interactions.

src/templates/partials/lighttpd.hbs
gstrauss: II will make the change as requested, but I politely disagree.  The tool gives recommendations for various security levels.  While the configs should not be noisy with explanations, I feel that it is beneficial to highlight serious security limitations.

src/templates/partials/lighttpd.hbs
gstrauss: Sorry to disappoint, but it does.  To avoid repetition, you can use the 'include' directive to include a separate file in each condition, and the separate file lists the ssl config once.

This is addressed a bit better in lighttpd 1.4.56.

src/templates/partials/lighttpd.hbs
gstrauss: ok.  comment removed.

src/templates/partials/lighttpd.hbs
gstrauss: (See #122 for how this is handled in lighttpd 1.4.56)

FYI: lighttpd 1.4.56 is on track to be released next weekend, so I am hoping you have a chance to review #122 soon, too!  Thanks!

gstrauss:
lighttpd 1.4.56 has been released.

Please let me know if you have further feedback here or in #122 

Thanks!

gstrauss:
Feedback here has incorporated and all comments have a response.

lighttpd 1.4.58 was released yesterday.

Please let me know if you have further feedback here or in #122

Thanks!

gstrauss:
Feedback here has incorporated and all comments have a response.

lighttpd 1.4.59 was released today.

Please let me know if you have further feedback here or in #122

Thanks!

gene1wood (approved):

gene1wood:
Thanks very much for the update @gstrauss

PR #123, opened 2021-01-26
Nginx: Wrap "return 301" inside a `location` block

### LINKED ISSUE #95: Would it better to wrap "return 301" inside a `location` block of nginx conf?
I've recently found that [acme.sh](https://github.com/acmesh-official/acme.sh) cannot auto renew a domain hosted by nginx and using the configure file generated by this tool. After some inspection, I found acme.sh is just adding a block like 

```
location ~ "^/\.well-known/acme-challenge/([-_a-zA-Z0-9]+)$" {
  default_type text/plain;
  return 200 "I AM THE TOKEN";
}
```

to the `listen 80` server block. However, cause the `return 301` is under the main `server` block, no other blocks won't override that.

I'm not sure if the best way is to make a change to acme.sh project, or conf template of this tool? From my point of view, default action seems to be always inside a `location /` block. But on the other side, it's acme.sh who should take more care to be compatible with more cases. Any ideas?

nmxcgeo:
This fixes #95.

I have tested the config in an nginx/1.14.2 on Debian. 
I have not checked if the code still builds, but I don't see, how modifying a template this way, would influence this.

Thanks to @runapp for the suggestion.

Greetings Nmxosm

### DIFF
--- a/src/templates/partials/nginx.hbs
+++ b/src/templates/partials/nginx.hbs
@@ -5,7 +5,9 @@ server {
     listen 80 default_server;
     listen [::]:80 default_server;
 
-    return 301 https://$host$request_uri;
+    location / {
+        return 301 https://$host$request_uri;
+    }
 }
 
 {{/if}}


gene1wood (approved):
This looks good and makes sense to me as an improvement.

PR #136, opened 2021-04-03
Document that minver returns true if equal

mavit:


gene1wood:
Good catch, `semver` uses [`gte`](https://github.com/mozilla/ssl-config-generator/blob/6871d64982ae26997a78d58630f50c203d6ea822/src/js/helpers/minver.js#L26) instead of `gt`.

Issue #117, opened 2020-09-29
308 Permanent Redirect


mavit:
I notice that 301 Moved Permanently is used to redirect from HTTP to HTTPS, but that this status code [can cause POST requests to be transformed to GET requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301).

There's a new code [308 Permanent Redirect](https://tools.ietf.org/html/rfc7538) which seems more appropriate, here.  I understand it's not supported by IE 11 before Windows 10, but is there a reason it's not used in the Modern profiles?

gene1wood:
Good point. [Looks like browser support](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308) would map to Modern.

Anyone interested in PRing changes for Modern in your favorite server type to use 308?

gstrauss:
According to the details in https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308, all web browsers listed on the page have supported 308 for over 8 years, with Edge, of course, being the last among them to add support for 308 in Edge 12 released 2015-07-28.

A year and a half ago, Microsoft published an article declaring IE 11 dead and buried.  https://blogs.windows.com/windowsexperience/2022/06/15/internet-explorer-11-has-retired-and-is-officially-out-of-support-what-you-need-to-know/
Again, that was a year and a half ago and was not a surprise, IE having been long deprecated.

Internet Explorer 11 desktop app retirement FAQ  (Published May 19 2021 08:55 AM)
https://techcommunity.microsoft.com/t5/windows-it-pro-blog/internet-explorer-11-desktop-app-retirement-faq/ba-p/2366549
In an update to that article on or before May 18, 2023:
> Update:
> The retired, out-of-support Internet Explorer 11 desktop application has been permanently disabled through a Microsoft Edge update on certain versions of Windows 10.
> Based on customer feedback, organizations will maintain control over when to remove IE11 UI elements from their devices.
> Over the coming months a small subset of exceptional scenarios where IE11 is still accessible will be redirected to Edge, ensuring users access a supported and more secure Microsoft browser.

https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions
Windows 8.1 end-of-life was 2023-01-10, a year ago.

---

IE has been *actively* disabled.  The concerns raised in this issue are out-of-date.

This issue could have been resolved years ago with a simple choice:
* Allow status code 308 to be used in templates without restriction
or
[...]

PR #137, opened 2021-04-03
For Apache, prefer 308 Permanent Redirect to 301 Moved Permanently

mavit:
Applies to the modern configuration only.

Relates to #117.


gstrauss:
lighttpd mod_redirect in lighttpd 1.4.75 and later takes a more nuanced approach by default and uses 301 Moved Permanently for all GET/HEAD requests or for HTTP/1.0 requests with other HTTP methods.  For HTTP/1.1 or later, lighttpd mod_redirect uses 308 Permanent Redirect for HTTP methods which are not GET/HEAD.

The Hypertext Transfer Protocol Status Code 308 (Permanent Redirect)  https://www.rfc-editor.org/rfc/rfc7238
was published June 2014, *over 10 years ago*.

For redirection involving HSTS, all clear-text http requests are redirected unconditionally to https.

Absent HTTP version support similar to what is in lighttpd 1.4.75, 308 Permanent Redirect should be used for all HSTS redirects, as that is the proper behavior for methods which are not GET/HEAD.  If a small number of end-users encounter problems with this recommended setting with specific old clients, then that small number of end-users can adjust the configs to use 301 instead of 308, and trade-off HSTS redirecting all HTTP methods other than GET/HEAD to HTTP GET method to the https location.

gstrauss:
@janbrasna you added a label that this change is blocked by the spec.  I understand that the implementation in this PR modifies the guidelines to add the permanent redirect code to the spec, but I do not think we should implement it that way.

If, in the future, we generate older versions of the site, we should do so from source control, or source control snapshots of the site generated at certain tags.  We are not currently testing that the current site including templates works with all prior versions of the guidelines, are we?  I don't see the utility of doing so.

gstrauss:
@mavit thank you for the PR.  As I am sure you understand, this PR from 2021 is somewhat stale.

Recent changes to the code added `output.hstsRedirectCode`, which is currently hard-coded to 301.  If in the future we use 308, that is what will change.

As we recently opened a discussion for updating the guidelines, I am also proposing using 308 for ssl-config-generator HSTS unconditional HTTP redirect from http:// to https://.  That may or may not be part of the guidelines, but definitely won't be part of guideline v5.6 modified by this PR, so I am going to close this PR.  We'll make changes pending the outcome of the discussions in #117.

Again, thank you for making the effort to submit this patch!

PR #151, opened 2021-06-19
Add Squid support

### LINKED ISSUE #112: Generate configurations for Squid
Hi,

It would be very useful if the config generator were extended to support [Squid](http://www.squid-cache.org/), a widely used HTTP proxy/cache. A list of SSL/TLS-related options is available under [http\_port](http://www.squid-cache.org/Doc/config/http_port/), as well as in the [config homepage](http://www.squid-cache.org/Doc/config/) under `sslproxy_`.

Thanks!

JGoutin:
The configuration is done to support Squid >= 3.5 ([The oldest version packaged in common Linux distributions](https://pkgs.org/download/squid))

The generated configuration was tested in the following conditions:
- Tested both "Intermediate" and "Modern" configurations
- Only in the "cache proxy with SSL bump" mode
- Only with Squid 4.14 on Fedora 34

I didn't find exactly what is the first Squid version that support TLS 1.3, so I set it to 3.5 in `configs.js`.

Close #112

### DIFF
--- a/src/js/configs.js
+++ b/src/js/configs.js
@@ -159,7 +159,7 @@ module.exports = {
   },
   squid: {
     highlighter: 'nginx',  // TODO: find better
-    latestVersion: '4.14',
+    latestVersion: '5.6',
     name: 'Squid',
     showSupports: false,
     supportsHsts: false,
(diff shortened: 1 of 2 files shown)


gene1wood:
Looks like [TLS 1.3 wasn't supported in Squid 3.5](http://www.squid-cache.org/Versions/v3/3.5/cfgman/sslproxy_version.html) so it must be Squid 4 that it starts

gene1wood:
Thanks @JGoutin !

PR #153, opened 2021-07-08
Remove useless matcher for caddy HSTS directive

rugk:
The matcher `/` for the Caddy header is not really necessary, as it is optional here.
The default for [matchers](https://caddyserver.com/docs/caddyfile/matchers) is the match *, i.e. every request.
Matching everything in the path `/` is basically the same (I tested it), so this change should make no difference.

See https://caddyserver.com/docs/caddyfile/directives/header


francislavoie:
Actually, the `/` matcher is harmful in this case, because it _only_ matches requests to exactly `/` and nothing else. Removing it makes it match _all_ requests which is what it _should_ do.

In Caddy v1, path matching was by prefix, but in v2, path matching is exact, so `/` is not correct.

rugk:
Okay, so this is an argument _in favor_ of this PR, is not it? :thinking: 

So it should match all requests, of course.

francislavoie:
Yes, definitely.

gene1wood:
@rugk Thanks for the fix.

@francislavoie I'm not that familiar with Caddy. The purpose of the 3 configuration profiles is to allow for a worse security posture with `Old` to be able to serve older clients while providing a better security posture with `Modern` if you don't need to support old clients.

Do you know how Caddy approaches this (balancing good security with support of older clients)? Would the default configuration you suggest support the `Old` clients like

Firefox 1, Android 2.3, Chrome 1, Edge 12, IE8 on Windows XP, Java 6, OpenSSL 0.9.8, Opera 5, and Safari 1

If you can share more detail on this in a new GitHub issue, we can work on figuring out how to best balance it.

gene1wood:
This is deployed

francislavoie:
Caddy v2 doesn't support lower than TLS 1.2 at all (because older TLS versions are completely broken). So all those clients won't work.

francislavoie:
IMO the "old" option for Caddy should be totally disabled. The intermediate option should remove all `tls` config (irrelevant) and turning off TLS 1.2 with "modern" is kinda silly and counterproductive, so I'd also just remove the `tls` config for that option as well.

Caddy's defaults are secure, there's no reason to tune cipher suites, and configuring cipher suites has no effect at all when using TLS 1.3 because the Go stdlib automatically ordering them. See https://go.dev/blog/tls-cipher-suites as I mentioned earlier.

janbrasna:
FWIW the Caddy config supports both Caddy v1.x and v2.x at the same time. So while the defaults are chosen wisely in Caddy v2.x, this tool still helps with configuring Caddy v1.x according to the Mozilla TLS recommendations.

Having safe defaults is great. This config generator doesn't asses any server version's defaults, but rather helps turning Mozilla's JSON recommendations of the three tiers into the server config. Nobody is forced to use them. They're just "recommendations", and reflect the currently published https://ssl-config.mozilla.org/guidelines/latest.json

francislavoie:
Caddy v1 is no longer supported, so it does not make sense at all to continue showing config for it. Don't recommend config for EOL software, please.

PR #175, opened 2022-07-28
Update dovecot DH based on new versions

### LINKED ISSUE #156: Dovecot 2.3.15 supports TLSv1.3
Please add modern config support for dovecot since v2.3.15. See https://doc.dovecot.org/settings/core/#setting-ssl-min-protocol.

Also dovecot does not require ssl_dh setting (or non-ECC DH) since v2.3.3, and can be left out. 

Thanks.

gene1wood:
https://doc.dovecot.org/configuration_manual/dovecot_ssl_configuration/#ssl-security-settings
Fixes #156

### DIFF
--- a/src/templates/partials/dovecot.hbs
+++ b/src/templates/partials/dovecot.hbs
@@ -5,8 +5,8 @@ ssl = required
 ssl_cert = </path/to/signed_cert_plus_intermediates
 ssl_key = </path/to/private_key
 {{#if output.usesDhe}}
-
-  {{#if (minver "2.3.0" form.serverVersion)}}
+  {{#if (minver "2.3.3" form.serverVersion)}}
+  {{else if (minver "2.3.0" form.serverVersion)}}
 # {{output.dhCommand}} > /path/to/dhparam
 ssl_dh = </path/to/dhparam
   {{else}}


src/templates/partials/dovecot.hbs
```diff
@@ -5,8 +5,8 @@ ssl = required
 ssl_cert = </path/to/signed_cert_plus_intermediates
 ssl_key = </path/to/private_key
 {{#if output.usesDhe}}
-
-  {{#if (minver "2.3.0" form.serverVersion)}}
+  {{#if (minver "2.3.3" form.serverVersion)}}
+  {{else if (minver "2.3.0" form.serverVersion)}}
 # {{output.dhCommand}} > /path/to/dhparam
 ssl_dh = </path/to/dhparam
```
janbrasna, line 11: @gene1wood Shouldn't this be left out only for `modern`? Since `intermediate` and `old` use DHE suites, they should still benefit from these dhparams(?)

janbrasna (commented):
> "Since v2.3.3+ Diffie-Hellman parameters have been made optional, and you are encouraged to disable non-ECC DH algorithms completely."
— [doc.dovecot.org/configuration_manual/dovecot_ssl_configuration](https://doc.dovecot.org/configuration_manual/dovecot_ssl_configuration/#ssl-security-settings)

This sounds to me as removing `ssl_dh` should go hand in hand with disabling `DHE-*` cipher suites, which only applies with `modern` configuration.

src/templates/partials/dovecot.hbs
janbrasna: Actually I think this should be reverted. Since it's already behind a `usesDhe` test (that's `true` only when any `DHE-*` is included in the output suites), it won't be output for `modern` where it's not needed, and for `intermediate` and `old` the 2.2.x vs. 2.3.x ways of `ssl_dh*` was correct.

janbrasna: What _might be needed_ is between v2.3–v2.3.3 adding the dhparams to the output always, unconditionally even when no DHE suites are present, as the docs state:

> "From version 2.3, you must specify path to DH parameters file using:"

(which is somewhat ambiguous, it's not clear from the docs that one must specify the path only if interested in own dhparams, or simply _always_…)

gene1wood: @janbrasna Can you open an issue on this so we can track it outside this PR and determine what change is needed? Maybe just copy paste in what you wrote above?

janbrasna: @gene1wood Worry not, will do once I have enough info on the last bit — I'm keeping it in mind.

PR #139, opened 2021-04-05
lighttpd: 308 permanent redirect

gstrauss:
lighttpd: 308 permanent redirect

depends on #137

PR #189, opened 2023-01-04
Use implicit settings for lighttpd ≥1.4.68

gstrauss:
Updates lighttpd to 1.4.76

Leaves out explicit settings when server defaults based on its version can be used implicitly.

- lighttpd 1.4.56 and later default to MinProtocol TLSv1.2
- lighttpd 1.4.68 and later default to a strict set of PFS ciphers and to -ServerPreference, since a strict set of PFS ciphers is the default.
- Simplify intermediate and modern configs for lighttpd 1.4.68 and later.
   _lighttpd defaults will incrementally be made more secure in the future and using lighttpd secure defaults (without explicitly hard-coding the defaults at a point in time) will allow users to get more secure defaults along with lighttpd upgrades, instead of accidentally continuing to use explicitly set older, less-secure config settings._

Adds future defaults that should be released in lighttpd with 1.4.78

### DIFF
--- a/src/templates/partials/lighttpd.hbs
+++ b/src/templates/partials/lighttpd.hbs
@@ -6,25 +6,6 @@
 #server.port = 80
 $SERVER["socket"] == "[::]:80" { }
 
-{{#if form.hsts}}
-$HTTP["scheme"] == "http" {
-{{#if (minver "1.4.50" form.serverVersion)}}
-    url.redirect = ("" => "https://${url.authority}${url.path}${qsa}")
-{{else}}
-    $HTTP["host"] =~ ".*" {
-        url.redirect = (".*" => "https://%0$0")
-    }
-{{/if}}
-}
-
-$HTTP["scheme"] == "https" {
-    # HTTP Strict Transport Security ({{output.hstsMaxAge}} seconds)
-    setenv.add-response-header = (
-        "Strict-Transport-Security" => "max-age={{output.hstsMaxAge}}"
-    )
-}
-{{/if}}
-
 {{#if (minver "1.4.56" form.serverVersion)}}
 # select one TLS module: "mod_openssl" "mod_mbedtls" "mod_gnutls" "mod_wolfssl" "mod_nss"
 #server.modules += ("mod_openssl")
@@ -38,18 +19,49 @@ ssl.privkey = "/path/to/private_key"
 ssl.pemfile = "/path/to/signed_cert_followed_by_intermediates"
  {{#if (minver "1.0.2" form.opensslVersion)}}
   {{#if (minver "1.1.0" form.opensslVersion)}}
-ssl.openssl.ssl-conf-cmd = ("MinProtocol" => {{#if (includes "TLSv1" output.protocols)}}"TLSv1"{{else if (includes "TLSv1.1" output.protocols)}}"TLSv1.1"{{else if (includes "TLSv1.2" output.protocols)}}"TLSv1.2"{{else}}"TLSv1.3"{{/if}})
+   {{#if (includes "TLSv1" output.protocols)}}
+ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1")
+   {{else if (includes "TLSv1.1" output.protocols)}}
+ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.1")
+   {{else if (includes "TLSv1.2" output.protocols)}}
+    {{#unless (minver "1.4.56" form.serverVersion)}}
+ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.2")
+    {{else}}
+#ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.2")  # lighttpd {{form.serverVersion}} TLS default
+    {{/unless}}
+   {{else}}
+ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.3")
+   {{/if}}
   {{else}}
 ssl.openssl.ssl-conf-cmd = ("Protocol" => "ALL, -SSLv2, -SSLv3{{#unless (includes "TLSv1" output.protocols)}}, -TLSv1{{/unless}}{{#unless (includes "TLSv1.1" output.protocols)}}, -TLSv1.1{{/unless}}{{#unless (includes "TLSv1.2" output.protocols)}}, -TLSv1.2{{/unless}}")
   {{/if}}
+  {{#if (minver "1.4.68" form.serverVersion)}}
+   {{#if output.serverPreferredOrder}}
+ssl.openssl.ssl-conf-cmd += ("Options" => "+ServerPreference")
+   {{/if}}
+  {{else}}
 ssl.openssl.ssl-conf-cmd += ("Options" => "{{#if output.serverPreferredOrder}}+{{else}}-{{/if}}ServerPreference")
+  {{/if}}
   {{#if output.ciphers.length}}
+   {{#if (minver "1.4.68" form.serverVersion)}}
+    {{#if (includes "old" form.config)}}
+ssl.openssl.ssl-conf-cmd += ("CipherString" => "{{{join output.ciphers ":"}}}")
+    {{else}}
+
+# lighttpd TLS defaults are widely supported by clients and should be preferred
+# See https://wiki.lighttpd.net/Docs_SSL
+# Uncomment to better match the less restricted Mozilla {{form.config}} spec.
+#ssl.openssl.ssl-conf-cmd += ("CipherString" => "{{{join output.ciphers ":"}}}")
+    {{/if}}
+   {{else}}
 # TLS modules besides mod_openssl might name ciphers differently
 # See https://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_SSL
 ssl.openssl.ssl-conf-cmd += ("CipherString" => "{{{join output.ciphers ":"}}}")
+   {{/if}}
   {{/if}}
   {{#if form.ocsp}}
-# OCSP stapling (input file must be maintained by external script)
+
+# OCSP stapling (input file must be maintained by external script, e.g. cert-staple.sh)
 # https://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_SSL#OCSP-Stapling
 ssl.stapling-file = "/path/to/cert-staple.der"
   {{/if}}
@@ -85,7 +97,7 @@ $SERVER["socket"] == ":443" {
     # {{form.config}} configuration
  {{#if (minver "1.4.48" form.serverVersion)}}
   {{#if (minver "1.1.0" form.opensslVersion)}}
-    ssl.openssl.ssl-conf-cmd = ("MinProtocol" => {{#if (includes "TLSv1" output.protocols)}}"TLSv1"{{else if (includes "TLSv1.1" output.protocols)}}"TLSv1.1"{{else if (includes "TLSv1.2" output.protocols)}}"TLSv1.2"{{else}}"TLSv1.3"{{/if}}, "Options" => "-SessionTicket")
+    ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "{{output.protocols.[0]}}", "Options" => "-SessionTicket")
   {{else if (minver "1.0.2" form.opensslVersion)}}
     ssl.openssl.ssl-conf-cmd = ("Protocol" => "ALL, -SSLv2, -SSLv3{{#unless (includes "TLSv1" output.protocols)}}, -TLSv1{{/unless}}{{#unless (includes "TLSv1.1" output.protocols)}}, -TLSv1.1{{/unless}}{{#unless (includes "TLSv1.2" output.protocols)}}, -TLSv1.2{{/unless}}", "Options" => "-SessionTicket")
   {{else}}
@@ -102,4 +114,22 @@ $SERVER["socket"] == ":443" {
  {{/if}}
 }
 #$SERVER["socket"] == "[::]:443" { ... } # repeat entire $SERVER["socket"] == ":443" { ... } config above for IPv6
+{{/if}}
+
+{{#if form.hsts}}
+$HTTP["scheme"] == "https" {
+    # HTTP Strict Transport Security ({{output.hstsMaxAge}} seconds)
+    setenv.add-response-header = (
+        "Strict-Transport-Security" => "max-age={{output.hstsMaxAge}}"
+    )
+}
+else $HTTP["scheme"] == "http" {
+ {{#if (minver "1.4.50" form.serverVersion)}}
+    url.redirect = ("" => "https://${url.authority}${url.path}${qsa}")
+ {{else}}
+    $HTTP["host"] =~ ".*" {
+        url.redirect = (".*" => "https://%0$0")
+    }
+ {{/if}}
+}
 {{/if}}
--- a/src/templates/partials/lighttpd.hbs
+++ b/src/templates/partials/lighttpd.hbs
@@ -21,13 +21,21 @@ ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1")
    {{else if (includes "TLSv1.1" output.protocols)}}
 ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.1")
    {{else if (includes "TLSv1.2" output.protocols)}}
-    {{#unless (minver "1.4.56" form.serverVersion)}}
+    {{#if (minver "1.4.78" form.serverVersion)}}
 ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.2")
     {{else}}
+     {{#unless (minver "1.4.56" form.serverVersion)}}
+ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.2")
+     {{else}}
 #ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.2")  # lighttpd {{form.serverVersion}} TLS default
-    {{/unless}}
-   {{else}}
+     {{/unless}}
+    {{/if}}
+   {{else if (includes "TLSv1.3" output.protocols)}}
+    {{#unless (minver "1.4.78" form.serverVersion)}}
 ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.3")
+    {{else}}
+#ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.3")  # lighttpd {{form.serverVersion}} TLS default
+    {{/unless}}
    {{/if}}
   {{else}}
 ssl.openssl.ssl-conf-cmd = ("Protocol" => "ALL, -SSLv2, -SSLv3{{#unless (includes "TLSv1" output.protocols)}}, -TLSv1{{/unless}}{{#unless (includes "TLSv1.1" output.protocols)}}, -TLSv1.1{{/unless}}{{#unless (includes "TLSv1.2" output.protocols)}}, -TLSv1.2{{/unless}}")
--- a/src/templates/partials/lighttpd.hbs
+++ b/src/templates/partials/lighttpd.hbs
@@ -60,14 +60,14 @@ ssl.openssl.ssl-conf-cmd += ("CipherString" => "{{{join output.ciphers ":"}}}")
     {{/if}}
    {{else}}
 # TLS modules besides mod_openssl might name ciphers differently
-# See https://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_SSL
+# See https://wiki.lighttpd.net/Docs_SSL
 ssl.openssl.ssl-conf-cmd += ("CipherString" => "{{{join output.ciphers ":"}}}")
    {{/if}}
   {{/if}}
   {{#if form.ocsp}}
 
 # OCSP stapling (input file must be maintained by external script, e.g. cert-staple.sh)
-# https://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_SSL#OCSP-Stapling
+# https://wiki.lighttpd.net/Docs_SSL#OCSP-Stapling
 ssl.stapling-file = "/path/to/cert-staple.der"
   {{/if}}
  {{else}}
--- a/src/templates/partials/lighttpd.hbs
+++ b/src/templates/partials/lighttpd.hbs
@@ -1,14 +1,11 @@
 # {{output.header}}
 # {{{output.link}}}
-#server.modules += ("mod_redirect")
-#server.modules += ("mod_setenv")
-#server.modules += ("mod_openssl")
 #server.port = 80
 $SERVER["socket"] == "[::]:80" { }
 
 {{#if (minver "1.4.56" form.serverVersion)}}
 # select one TLS module: "mod_openssl" "mod_mbedtls" "mod_gnutls" "mod_wolfssl" "mod_nss"
-#server.modules += ("mod_openssl")
+server.modules += ("mod_openssl")
 
 # lighttpd 1.4.56 and later will inherit ssl.* from the global scope if
 # $SERVER["socket"] contains ssl.engine = "enable" and no other ssl.* options
@@ -74,6 +71,9 @@ ssl.cipher-list = "{{{join output.ciphers ":"}}}"
   {{/if}}
  {{/if}}
 {{else}}
+ {{#if (minver "1.4.46" form.serverVersion)}}
+#server.modules += ("mod_openssl")
+ {{/if}}
 $SERVER["socket"] == ":443" {
     ssl.engine  = "enable"
 
@@ -117,6 +117,13 @@ $SERVER["socket"] == ":443" {
 {{/if}}
 
 {{#if form.hsts}}
+ {{#if (minver "1.4.56" form.serverVersion)}}
+server.modules += ("mod_redirect")
+server.modules += ("mod_setenv")
+ {{else}}
+#server.modules += ("mod_redirect")
+#server.modules += ("mod_setenv")
+ {{/if}}
 $HTTP["scheme"] == "https" {
     # HTTP Strict Transport Security ({{output.hstsMaxAge}} seconds)
     setenv.add-response-header = (
--- a/src/js/configs.js
+++ b/src/js/configs.js
@@ -97,7 +97,7 @@ module.exports = {
   },
   lighttpd: {
     highlighter: 'nginx',
-    latestVersion: '1.4.67',
+    latestVersion: '1.4.76',
     name: 'lighttpd',
     tls13: '1.4.48',
   },


gstrauss:
@gene1wood please review.  Thank you.

This PR was submitted in Jan for lighttpd 1.4.68, updated in Feb for lighttpd 1.4.69, and now once again updated in May for lighttpd 1.4.70.

gstrauss:
@gene1wood is there a chance that this could get reviewed?  It is a year old.

src/templates/partials/lighttpd.hbs
```diff
@@ -38,18 +19,40 @@ ssl.privkey = "/path/to/private_key"
...
+ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1")
+   {{else if (includes "TLSv1.1" output.protocols)}}
+ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.1")
+   {{else if (includes "TLSv1.2" output.protocols)}}
+    {{#if (minver "1.4.56" form.serverVersion)}}{{else}}
+ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.2")
+    {{/if}}
+   {{else}}
+ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.3")
+   {{/if}}
```
gene1wood, line 32: @gstrauss It seems like this logic doesn't work. When I check this PR out and run it, selecting `Intermediate` and lighthttpd version `1.4.67` I get

```
...
ssl.pemfile = "/path/to/signed_cert_followed_by_intermediates"
ssl.openssl.ssl-conf-cmd += ("Options" => "-ServerPreference")
# TLS modules besides mod_openssl might name ciphers differently
...
```

And the `ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.2")` line is missing.

I'm unsure of the logic you intended.

The `Intermediate` configuration has a `output.protocols` list of `["TLSv1.2", "TLSv1.3"]`

* What `MinProtocol` should lighthttpd have for `Intermediate` on version `1.4.67`?

Also if you change the `{{#if (minver "1.4.56" form.serverVersion)}}{{else}}` line to use `unless` instead of `if` with an empty `else` it will fix the formatting issue that causes the line to be preceded by 4 spaces causing it to look like

```
    ssl.openssl.ssl-conf-cmd += ("Options" => "-ServerPreference")
```

Here's what it would look like with `unless`

```
    {{#unless (minver "1.4.56" form.serverVersion)}}
ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.2")
    {{/unless}}
```

gene1wood (changes requested):

src/templates/partials/lighttpd.hbs
gstrauss: `"MinProtocol" => "TLSv1.2"` is the default in lighttpd TLS modules since lighttpd 1.4.56.
It is preferred that people use lighttpd's secure defaults -- which are periodically reviewed, announced, and updated -- rather than specifying something which may then be cargo-culted for years and never revisited in user configurations.

I'll make the suggested `unless` change shortly and will push an updated branch rebased on master.

gstrauss: requested changes have been made

gstrauss: FYI, the commit message for a56011b7 in this PR explains:
```
lighttpd updates for lighttpd 1.4.68

lighttpd 1.4.56 and later default to MinProtocol TLSv1.2

lighttpd 1.4.68 and later default to a strict set of PFS ciphers and
to -ServerPreference, since a strict set of PFS ciphers is the default.

Simplify intermediate and modern configs for lighttpd 1.4.68 and later.
lighttpd defaults will incrementally be made more secure in the future
and using lighttpd secure defaults (without explicitly hard-coding the
defaults at a point in time) will allow users to get more secure
defaults along with lighttpd upgrades, instead of accidentally
continuing to use explicitly set older, less-secure config settings.
```

gstrauss: Besides rebasing this PR onto tip of current master branch, the only substantive change in the push below is your requested change to use `unless` in `src/templates/partials/lighttpd.hbs`

janbrasna: Not sure if this should include 308 redirects, that ought to be addressed separately #117 — there were discussions in #137 regarding browser support and selective addition to only modern profile etc.; there was also impact to .json specs so that would need to get versioned etc. …

src/templates/partials/lighttpd.hbs
```diff
@@ -38,18 +19,40 @@ ssl.privkey = "/path/to/private_key"
...
   {{else}}
 ssl.openssl.ssl-conf-cmd = ("Protocol" => "ALL, -SSLv2, -SSLv3{{#unless (includes "TLSv1" output.protocols)}}, -TLSv1{{/unless}}{{#unless (includes "TLSv1.1" output.protocols)}}, -TLSv1.1{{/unless}}{{#unless (includes "TLSv1.2" output.protocols)}}, -TLSv1.2{{/unless}}")
   {{/if}}
+  {{#if (minver "1.4.68" form.serverVersion)}}
+   {{#if output.serverPreferredOrder}}
+ssl.openssl.ssl-conf-cmd += ("Options" => "+ServerPreference")
+   {{/if}}
+  {{else}}
 ssl.openssl.ssl-conf-cmd += ("Options" => "{{#if output.serverPreferredOrder}}+{{else}}-{{/if}}ServerPreference")
+  {{/if}}
```
janbrasna, line 42: Is it worth differentiating? The JSON specs say either set preference, or don't set preference. Is it better to just omit it and leave to server default?

src/templates/partials/lighttpd.hbs
```diff
@@ -38,18 +19,40 @@ ssl.privkey = "/path/to/private_key"
...
   {{#if output.ciphers.length}}
+   {{#if (minver "1.4.68" form.serverVersion)}}
+    {{#if (includes "old" form.config)}}
+ssl.openssl.ssl-conf-cmd += ("CipherString" => "{{{join output.ciphers ":"}}}")
+    {{/if}}
+   {{else}}
 # TLS modules besides mod_openssl might name ciphers differently
 # See https://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_SSL
 ssl.openssl.ssl-conf-cmd += ("CipherString" => "{{{join output.ciphers ":"}}}")
+   {{/if}}
```
janbrasna, line 52: This means not setting the cipher suites since ≥v1.4.68 unless `old`, i. e. `modern` and `intermediate` won't set the suites at all according to json specs? Isn't the intention of the configs to explicitly set the cipher suites according to the specs, instead of leaving it to the server defaults?

janbrasna (commented):
I'm not sure I understand the changes either, could you please confirm the meaning behind:

src/templates/partials/lighttpd.hbs
janbrasna: (I mean, the original +/- satisfied the specs exactly. Is there a reason starting ≥v1.4.68 for not setting `-ServerPreference` explicitly?)

src/templates/partials/lighttpd.hbs
janbrasna: @gene1wood or do I read it wrong? I understand there are more and more great defaults provided by the servers, but that doesn't mean the generator should give up and not output the cipher suites according to the current .json recommendations (which is whole another discussion TBH).

Is there a consensus when the actual output should be left out for server defaults behind a version conditional?

src/templates/partials/lighttpd.hbs
gstrauss: @janbrasna the commit message cut-n-paste into the original description of this PR submission explains that it is best to use the lighttpd TLS defaults.  You're welcome to disagree and I am happy to discuss further.  However, I do not see how repeating that text a third time in this PR would be useful when I think that text already answers your question.

> Is it better to just omit it and leave to server default?

Yes.  I am a lighttpd developer and answer user questions on lighttpd forums.  Yes, it is better to use lighttpd defaults.

src/templates/partials/lighttpd.hbs
gstrauss: @janbrasna as above, yes, it is better to use lighttpd secure TLS defaults.

gstrauss: @janbrasna if you have further questions how to configure lighttpd securely, please refer to the [lighttpd TLS docs](https://wiki.lighttpd.net/Docs_SSL)

src/templates/partials/lighttpd.hbs
gstrauss: @janbrasna have you looked at https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308 recently?  All modern, supported browser clients support 308.  If you have a specific concern, please be more specific.  "Fear, Uncertainty, and Doubt" absent any **real data** are not strong arguments or productive contributions to a conversation.

[RFC7538](https://www.rfc-editor.org/rfc/rfc7538) "The Hypertext Transfer Protocol Status Code 308 (Permanent Redirect)" was published in 2015, **almost 9 years ago**.

gene1wood: @gstrauss Thanks for the `unless` fix to the indentation.

> `"MinProtocol" => "TLSv1.2"` is the default in lighttpd TLS modules since lighttpd 1.4.56.

Got it, what I meant is that it looks like the logic you wrote is intended to output the line

```
ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.2")
```

However it doesn't.

Even with the default you mention, will the subsequent line

```
    ssl.openssl.ssl-conf-cmd += ("Options" => "-ServerPreference")
```

work given that it's using a `+=` which I imagine concatenates onto an existing string? I ask because in the code's current state there is no existing string to add to.

Here's the output when I run this PR's code and choose lighthttp 1.4.67

image

Ya, I'd been intending to try to work on the HTTP 308 topic distinct from this PR, but don't worry about that, I can fix this up. Once we have the logic for the `MinProtocol` fixed I'll fix things up and add more comments to #117

gstrauss: > Even with the default you mention, will the subsequent line
>
>    `ssl.openssl.ssl-conf-cmd += ("Options" => "-ServerPreference")`
>
> work given that it's using a += which I imagine concatenates onto an existing string? I ask because in the code's
> current state there is no existing string to add to.

@gene1wood: Yes, it will work as intended.

gene1wood: Ok cool, ya appending to empty lists like this was added in [1.4.17](https://redmine.lighttpd.net/versions/8) as you say.

janbrasna: @gstrauss No need to worry about me, I work with MDN daily. If pointing to previous maintainers' decisions (as a reason why more related HTTP 308 issues/PRs are still hanging around unresolved) feels like a FUD then I'm sorry, but I felt co-authors' decisions were more relevant, than e. g. my recollections of why this hasn't moved forward. That said… I just tried to explain that separating the 308 feature into its own might help expedite the proposed changes.

(There's no consensus on 308s and once decided, that should be projected into all server templates, behind their appropriate version conditionals i. e. ~ sure I'd love to see them given the POST handling, but historically there was the question if they belong to `intermediate` where IE11@W7 is buggy and `old` where most of the UAs claimed to be compatible can't redirect it. That's why I wanted to move this discussion to its own issue, and iron out the details there.)

src/templates/partials/lighttpd.hbs
janbrasna: @gstrauss I sure read it… and expected the defaults to be unnecessary old protocols and such, not completely omitting @mozilla/server-side-tls's `intermediate` cipher suites ;D that's why I raised the additional questions if I read it correctly and the resulting output is really meant that way and not a mistake.

gstrauss:
@gene1wood please consider using `git rebase` instead of `git merge` into contributor branches.  The git history will be much clearer.  In particular, resolving merge conflicts, if any, have the proper focus.  A conflict with `git merge` involves merging changes made elsewhere (of wide and varied nature) into the contributor branch.  A conflict with `git rebase` involves resolving conflicts with changes made specifically in the contributor branch.

On the contributor branch: `git fetch; git rebase origin/master` rather than `git fetch; git merge origin/master`

src/templates/partials/lighttpd.hbs
janbrasna: @gstrauss Nope, I don't have further questions how to configure lighttpd securely, I don't think I need to. Surprisingly that's not the issue.

My question is about this change, that basically disallows the output of `https://statics.tls.security.mozilla.org/server-side-tls-conf-*.json` and leaves that to the vendor's definition. While claiming it's the @mozilla/server-side-tls's e. g. `intermediate` … when it is no longer so;)
_(even if the effectively used suites match today, that might not be true tomorrow — and that's why Mozilla's TLS specs are versioned and explicit, to be predictable, if needed for any given usecase e. g. PCI DSS testing etc.)_

This PR basically reads for `intermediate` and `latestVersion` i. e. the default in the UI: "Claim it's Mozilla's https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28recommended.29 while in reality supplying Lighty's https://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_SSL#Perfect-Forward-Secrecy-PFS …"

There are more vendors wishing the output not to mess with their secure TLS defaults so I'll open a broader discussion for it. I understand it, servers with great defaults should get their credit for having safe defaults, yet I don't see the need to cripple the json specs output to achieve this.

janbrasna: Quickly comparing the lists, lighty's includes CCM/8s & PSKs for embeds that are not too important here, but misses DHE-RSA-GCMs from server-side-tls-conf-* specs.

So the end output claiming to be @mozilla intermediate is in reality not that:

Screen Shot 2024-01-05 at 23 11 31

See I'm no @april to have a say in this, I'm just testing the rendered output — and with no `configurations.intermediate.ciphers[]` present in the output at all, in the current state of affairs, that seemed like a bug. That's why I raised it for others to decide.

The need to somehow endorse great defaults where applicable is a whole another topic, unrelated to keeping the @mozilla's intermediate config output for lighty still to @mozilla's specs;)

gstrauss: @janbrasna as you mentioned already, the mozilla TLS recommendation specification is probably due to for review, and I agree.

The last substantial updates to the ssl-config-generator specs were in 2020, when april published versions 5.3, 5.4, 5.5, and 5.6 (July 2020), **now 3 1/2 years ago**.  Gene published 5.7 in May 2023, with the simple change to add DHE-RSA-CHACHA20-POLY1305.

I am not suggesting whether or not dhe-rsa-aes128-gcm-sha256 or dhe-rsa-aes256-gcm-sha384 should be included.  I am saying that the spec should be reviewed to see if they should, or should not be removed from the intermediate cipher list.  Once that is refreshed (and not 3 1/2 years old), then we can have a more relevant conversation about whether or not ssl-config-generator should emit a cipher list to include those in the ssl-config-generator "Intermediate" lighttpd config.

One data point to compare with others: the current Cloudflare recommendations include only ciphers using elliptic curve diffie hellman ephemeral (ECDHE) for key exchange in both their "Modern" and "Compatible" cipher lists.  https://developers.cloudflare.com/ssl/reference/cipher-suites/recommendations/

Another data point is that I have not received any complaints about lighttpd TLS defaults since lighttpd updated its TLS defaults in Jan 2023 (after having announced the scheduled change throughout all of 2022).  I have received zero requests for lighttpd to include those dhe ciphers in the lighttpd TLS defaults, and I do not have any plans to add them.

gstrauss: https://github.com/mozilla/server-side-tls/issues/285 notes "Note that DHE suites in TLS 1.2 are problematic." and discusses why Gene added DHE-RSA-CHACHA20-POLY1305 in ssl-config-generator 5.7.json

gstrauss:
To hopefully keep this PR fresh and unblocked, I rebased this PR on the master branch and added a commit to emit 308 redirect only for Intermediate and Modern configs, leaving the existing lighttpd default redirect behavior (sending 301 for redirects) for the Old config.
```
--- a/src/templates/partials/lighttpd.hbs
+++ b/src/templates/partials/lighttpd.hbs
@@ -115,9 +115,11 @@ $HTTP["scheme"] == "https" {
     )
 }
 else $HTTP["scheme"] == "http" {
+ {{#unless (includes "old" form.config)}}
  {{#if (minver "1.4.31" form.serverVersion)}}
     url.redirect-code = 308
  {{/if}}
+ {{/unless}}
  {{#if (minver "1.4.50" form.serverVersion)}}
     url.redirect = ("" => "https://${url.authority}${url.path}${qsa}")
  {{else}}
```
@gene1wood: If that is still a blocker for you, please remove `url.redirect-code` from the lighttpd config and use the lighttpd defaults.

@janbrasna I updated #117 with links that note that Windows versions before Windows 10 are all end-of-life and no longer receiving security updates, and IE in Windows 10 is end-of-life and no longer receiving security updates, so the any concern or questions about widespread support for 308 status code should have been resolved over a year ago.

gstrauss:
@gene1wood: ping

Are there remaining questions or feedback preventing this PR from moving forwards?

gene1wood:
@gstrauss I will try to set aside time in the next few days to look at this.

gstrauss:
@gene1wood: ping

Are there remaining questions or feedback preventing this PR from moving forwards?

gstrauss:
@gene1wood: ping

Are there remaining questions or feedback preventing this PR from moving forwards?

PR has been rebased on to tip of ssl-config-generator master branch and updated for lighttpd 1.4.74.

gstrauss:
@gene1wood: ping

Are there remaining questions or feedback preventing this PR from moving forwards?

---

With the release of lighttpd 1.4.75, the discussion over the almost **!!! 3 year old !!!** 308 Permanent Redirect issue #137 for Apache, and subsequent #139 for lighttpd is mostly moot for lighttpd since lighttpd 1.4.75 mod_redirect by default uses 308 for HTTP/1.1 (and later) requests, though lighttpd 1.4.75 still defaults to use 301 for HTTP/1.0 requests.  This PR does not add `url.redirect-code` for lighttpd 1.4.75 and later, which is the same as the current behavior in ssl-config-generator.

To avoid sending information in query strings, POST is often used, so using 308 instead of 301 is important for proper security postures when URL redirects are necessary to maintain compatibility with site changes.  It is a sad state of affairs that #137 has not been resolved in almost 3 years (!!!) with usage/transition guidance though modern, intermediate, and old configs.

gstrauss:
@gene1wood: ping

Are there remaining questions or feedback preventing this PR from moving forwards?

gstrauss:
@gene1wood: ping

Are there remaining questions or feedback preventing this PR from moving forwards?

gstrauss:
To summarize the above points of potential contention:

Regarding cipher lists, I previously wrote:
> I am not suggesting whether or not dhe-rsa-aes128-gcm-sha256 or dhe-rsa-aes256-gcm-sha384 should be included. I am saying that the [Mozilla] spec should be reviewed to see if they should, or should not be removed from the intermediate cipher list. Once that is refreshed (and not 3 1/2 years old), then we can have a more relevant conversation about whether or not ssl-config-generator should emit a cipher list to include those in the ssl-config-generator "Intermediate" lighttpd config.

The mozilla spec is now over 4 years old.  I will commit to updating lighttpd.hbs to match the Mozilla specs as soon as the Mozilla specs are updated.  @janbrasna, @gene1wood if you feel strongly that those ciphers should be included in the lighttpd intermediate config, please let's continue the conversation.

Regarding HSTS redirects with 308 Permanent Redirect, I previously wrote:

> With the release of lighttpd 1.4.75, the discussion over the almost !!! 3 year old !!! 308 Permanent Redirect issue https://github.com/mozilla/ssl-config-generator/pull/137 for Apache, and subsequent https://github.com/mozilla/ssl-config-generator/pull/139 for lighttpd is mostly moot for lighttpd since lighttpd 1.4.75 mod_redirect by default uses 308 for HTTP/1.1 (and later) requests, though lighttpd 1.4.75 still defaults to use 301 for HTTP/1.0 requests. This PR does not add url.redirect-code for lighttpd 1.4.75 and later, which is the same as the current behavior in ssl-config-generator.

In this PR, the change in spec produced by ssl-config-generator is for lighttpd 1.4.31 - lighttpd 1.4.74 to add `url.redirect-code = 308`, but is once again removed (the same as current spec) for lighttpd 1.4.75 - lighttpd 1.4.76, the current release of lighttpd.  @janbrasna if you have concerns, then I will remove the `url.redirect-code = 308`, as most users coming to ssl-config-generator for the first time are hopefully configuring the current version of lighttpd, where this is a non-issue.

https://repology.org/project/lighttpd/badges?exclude_unsupported=1 shows that the *current* releases of many popular distros are running lighttpd 1.4.76.

gstrauss:
@gene1wood you are currently listed as reviewer on this PR.  To give you a chance to comment, if you like, I'll wait another week before removing you as reviewer.  I then plan to merge this PR.  Thanks!

janbrasna:
I've deliberately changed the title to better reflect the content — and will look into the individual changes and the complexity vs. benefits they add.

I originally envisioned this to be split into more actionable chunks if some of the changes are blocked on the specs, but will look into the formal compatibility and think about adding some comments to let the consumers know if the declared support profile from the sstls json can't be satisfied 100%.

The proposed logic is asking for another lighttpd author to review tbh;)

gstrauss:
> The proposed logic is asking for another lighttpd author to review tbh;)

lol.  Please consider reviewing the commits individually.  They are somewhat smaller.

I'm willing to create further smaller commits based on your feedback, and am interested in moving this forwards so that new users to this site benefit from recently updated recommendations, rather than outdated, years-old recommendations.

gstrauss:
I am going to push a change so that for lighttpd 1.4.68 and later, for Intermediate config, the ciphers are listed, but commented out:
> \# lighttpd TLS defaults are widely supported and should be preferred
> \#ssl.openssl.ssl-conf-cmd += ("CipherString" => "ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:DHE-RSA-CHACHA20-POLY1305")

gstrauss:
> The proposed logic is asking for another lighttpd author to review tbh;)

I am the sole author of the TLS modules in lighttpd since lighttpd 1.4.56 when I rewrote the older lighttpd mod_openssl code to openssl 1.0.2 interfaces and subsequently added support for openssl 1.0.0, 1.1.0, and 3.0.0.  In lighttpd 1.4.56, I also wrote the lighttpd modules supporting alternative TLS libraries: GnuTLS, mbedTLS, wolfSSL, and NSS.

If you have more specific questions, I'll certainly pass them along to stbuehler, but I am not sure it is a good use of his time to bother him to wade into HandlebarJS and how its limitations and its use in ssl-config-generator logic leads to long, complex code conditions and duplication of results between various conditional statements.

Modern lighttpd still works with very old OpenSSL (though not recommended for security reasons), and supporting the matrix of many openssl and many lighttpd versions unfortunately leads to many conditions.

janbrasna:
Yes, commented out with a reasoning info might be the way to expose that. (But would like to discuss in a broader meeting first.)

Please note this is currently not broken or buggy, so the priority is lower than some of the bugfixes and compatibility updates that will need more attention first (quite the opposite, the new edge cases this brings need to be documented etc., so it actually introduces new known issues to tiptoe around…) — I think we should do a review meeting, or raise the more "meta" questions, that shouldn't be mixed up with feature bumps, beforehand, though. I will be able to address these within the next _weeks_, not _hours_ or _days_.

@gstrauss When you have a chance, please squash any version bumps that don't include any actual output logic impact into the next feature commit, trying to fixup as much of the noise for easier review of the whole timeline… Thanks.

janbrasna:
> > The proposed logic is asking for another lighttpd author to review tbh;)
> 
> lol. Please consider reviewing the commits individually. They are somewhat smaller.

They certainly include more rationale individually, with explainers in commit messages, but still as a whole this brings so much tribal knowledge necessary to understand what's the reason for every single condition, that nobody else would be able to maintain it;)

I'm looking into some reasonable hbs comments (i.e. not published to output, but describing the template code), but they do seem to muck up the output whitespace, so I had not much luck trying them out…

gstrauss:
> @gstrauss When you have a chance, please squash any version bumps that don't include any actual output logic impact into the next feature commit, trying to fixup as much of the noise for easier review of the whole timeline… Thanks.

squashed and rebased on tip of master

> Please note this is currently not broken or buggy, so the priority is lower than some of the bugfixes and compatibility updates that will need more attention first

It is years out-of-date, so should be considered buggy, similar to providing advice for EOL openssl versions and not providing the best advice for supported openssl 3.x versions (#256).  Yes, broken parts of the site should get priority.  Then buggy issues immediately after.

gstrauss:
> I'm looking into some reasonable hbs comments (i.e. not published to output, but describing the template code), but they do seem to muck up the output whitespace, so I had not much luck trying them out…

Have you tried HTML comments?  `<!-- ... -->`  It would be better not to bloat the pages, but appropriate comments are better than none.

src/templates/partials/lighttpd.hbs
janbrasna: Yes, for grading/testing consistency. Which is my point exactly with almost any similar change. If we promise a config tier, and a tool down the road uses that same definition to probe such config, and we fall short of the promise, that sucks.

E.g.. from another thread:

> "I humbly request that we prioritize updating this site to be at least as strict as the specifications, but stricter is OK."

Is presence of any CCM8 "stricter", or "at least as strict"? Who decides that? (RHEL folks would say "no", FWIW…)

So to avoid similar discussions, unless the usage of the tool warrants such one-off deviation from the recommendation data (think known issues with interop, key size or rotation etc.), I'd prefer not to challenge that HERE, but be helpful upstream, in defining what should the recommended profiles be, and align with other security folks from other domains adding another views to the issue as well (it's not all http servers and web browsers…). More secure defaults are only better for everyone, and pointing that out via comments is a good way to show no fine-tuning is really needed — but at the same time, however you may dislike the Mozilla TLS recommendations, this is only to turn them consistently into a config example, as little opinionated as possible TBH.

If you want to add a band-aid over the recommendations you don't agree with as a server author, it should be clearly stated though (for visibility, and to find a way folks are aware of).

gstrauss:
Added comment to the template:
```
 # lighttpd TLS defaults are widely supported by clients and should be preferred
 # See https://wiki.lighttpd.net/Docs_SSL
+# Uncomment to better match the *less secure* Mozilla {{form.config}} spec.
 #ssl.openssl.ssl-conf-cmd += ("CipherString" => "{{{join output.ciphers ":"}}}")
```

src/templates/partials/lighttpd.hbs
gstrauss: > however you may dislike the Mozilla TLS recommendations

That *is not* the issue.  The Mozilla TLS recommendations are outdated and currently appear to be unmaintained.  They have been supplanted by more up-to-date and maintained security recommendations from other qualified professional organizations.  One example: Cloudflare CDN https://developers.cloudflare.com/ssl/edge-certificates/additional-options/cipher-suites/recommendations/

As such, the Mozilla TLS recommendations, currently over 4 years old, should be reviewed and updated by Mozilla.  Until then, they should be discarded and ignored in favor of current and maintained TLS recommendations from other qualified professional organizations.  This is my professional opinion as a systems security programmer.  I would defend this statement to any CISO, CIO, or CTO if any project manager tried to pull "but look at *this* spec!"

gstrauss:
> > I'm looking into some reasonable hbs comments (i.e. not published to output, but describing the template code), but they do seem to muck up the output whitespace, so I had not much luck trying them out…
> 
> Have you tried HTML comments? `<!-- ... -->` It would be better not to bloat the pages, but appropriate comments are better than none.

Some other `src/templates/partials/*.hbs` templates use `{{! ... }}`.  Have you tried that?  Also, Handlebars has `~` modifier on the directive after the `{{` to affect whitespace of template directives.  Still, using a real programming language like javascript gives explicit control of the whitespace, instead of having to fight with the templating framework Handlebars, which is built on top of javascript.

src/templates/partials/lighttpd.hbs
gstrauss: I have removed the 308 changes from this PR.  First, the lighttpd 1.4.75 was released over 7 months ago and does something better by default.  Second, I'm willing to continue the 308 discussion knowing that I can push for a decision/resolution (one way or another) in a reasonable timeframe, e.g. within a couple months.

gstrauss:
> Merging is blocked
> Merging can be performed automatically once the requested changes are addressed.

eh, seems like I have to work around github rules here since it won't let me merge until one of the reviewers approves, so I am going to add myself as a review and approve.

gstrauss:
@gene1wood it appears that I can no longer add myself as a reviewer, which I was able to do to other PRs.  Perhaps it is because I created this PR?  Or maybe it is because reviewers were already assigned when you changed the repository settings?  Either way, I am unable to proceed here.  There are still rules in place on github which prevent me from pushing these changes to master.  Would you please look into the repository settings?  Or we can look at them together at on Tue meeting.

janbrasna:
There are still some default branch protections in place, and requested changes may still be one of them (in this case from someone with a higher role than those trying to dismiss them) which sounds reasonable. (And no, you can't be reviewer of your own PRs/pushes no matter how the checks are set up…) — I'll see if my review would be enough to allow this to merge once I go over the current changes.

The 308 approach in Lighty is lovely (i.e. will secure and preserve posted data, while still allowing access to the GET / index for clients not understanding 308 at all) and is something very carefully reasonable which I'm happy with. (I'll need to check what's the actual history of changes for that, if that current version default behaves with the same graceful degradation logic as explicit code set in older versions, or that would be a wholesale change and just the defaults are that smart etc. — and wether to explicitly include not-so-sensible ballpark setting for older versions or not… will probably go over those finicky details at once in a review meeting at some point.)

src/templates/partials/lighttpd.hbs
janbrasna: @gstrauss You're happy to ignore them as you wish, and also to work on any other organization's docs and tools (or your forks thereof) in the meantime… and I'll be happy to continue reviewing this changeset once the upstream config data match your expectations so you don't have to hardcode your disagreement in lieu of having to vandalize the outputs. (I'm perfectly fine with noting that in the output tho with the matching config included for reference for those envs that may need an exact match, so I'll look into that once I have some time to test drive a build with some of the recent changes…)

janbrasna: But the more important bit is — you'd probably need to do some of your own testing to make sure at least the publicly available test suites, when comparing to "Mozilla intermediate" or similar profile for OK/NOK or grading, won't complain about CCM8 and PSK suites that you include extra.

(I don't know honestly. They are all sorta specific use, not sure I'd be recommending them for general use without any more explanation — and not sure if any e.g. PCI testing suite wouldn't complain, for reasons unknown to me TBH, such as the plaintext username transfer in "pure" PSK for ≤TLS1.2 perhaps?)

@gstrauss How do you expose these? Are they enumerated to be explicitly enabled, or do you use some addition/subtraction based on defaults, so if they were not in default strings on the OS level, you're not adding them anyway…? Or…?

janbrasna: Read: If you need those, you need to know what to do and also enable them on purpose. We don't cater for such groups here in the configs. I'm not sure what all needs to be configured for them to even be able to end up negotiated — I'm only concerned about any automation/testing results that may stem from diverging from the exact restricted set in the jsons (as can be observed in the https://github.com/mozilla/server-side-tls/issues/285 rationale — following any allowlists perhaps too literally…)

janbrasna: The omission of DHE-RSA-AES* is another topic, don't remember if just the DHE-RSA-CHACHA20 you provide is enough to support the sad SChannel compatibility when we had to keep the DHEs back in the day for RSA-only compatibility… will look into those, but quickly glancing https://learn.microsoft.com/en-us/windows/win32/secauthn/tls-cipher-suites-in-windows-7 it seems that would not support the oldest clients we claim unless ECDSA is used 🤷

janbrasna: _(and this is the exact reason I'm not willing to burn too much time making the configs "too implicit" and having to hunt for compatibility woes one by one… when we have a set that makes the conservative compatibility match the guidelines exactly, unless noted otherwise in the comments)_

src/templates/partials/lighttpd.hbs
```diff
@@ -38,19 +16,54 @@ ssl.privkey = "/path/to/private_key"
...
+    {{else}}
+     {{#unless (minver "1.4.56" form.serverVersion)}}
+ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.2")
+     {{/unless}}
+    {{/if}}
+   {{else if (includes "TLSv1.3" output.protocols)}}
+    {{#unless (minver "1.4.78" form.serverVersion)}}
+ssl.openssl.ssl-conf-cmd = ("MinProtocol" => "TLSv1.3")
+    {{/unless}}
+   {{/if}}
```
janbrasna, line 35: (This is one of the extremely complicated ways to "not emit defaults" at any cost, making the actual conditional spaghetti hard to maintain and understandable without some tribal knowledge. I'd much rather just print `output.protocols[0]` as the lowest version /sorted/ and be done. TBD… Might be OK~ish tentatively if addressed once more powerful templating logic is available.)

src/templates/partials/lighttpd.hbs
janbrasna: I'm not really able to follow all the changes to Lighty defaults over time and with versions (hence why I'm hesitant to leave things to defaults and monitor changes constantly) but it seems that since the above comparison was posted, it's a bit stricter now — @gstrauss do I read it right that the PSK and DHE-RSA-CHACHA20 is gone now? And the presence of CCM/CCM8 is based on whether it was available in defaults before configuring the context? That would make the 3/24+ defaults reasonable, and only a note mentioning the absence of DHE thus the need for ECDSA chain to support the actual profile would suffice.

(BTW the actual Lighty cipher defaults are being documented under #PFS heading, it might be worth linking that section directly from the output comment as that's a valuable reference somewhat hard to find in general — as to why it's actually commented out, for comparison…)

**The main benefit of not configuring the cipherstring explicitly would be enabling compatibility with other TLS libraries — and that's a very important improvement!**

gstrauss:
@april thank you for the history.

As an experiment, last week I rewrote all the .hbs partials into javascript in a development branch.  The result saves over 70 KB (!) size in javascript over the Handlebars-generated javascript and also allows some shared code as well as simplification of conditions since javascript supports multiple conditions (e.g. `&&` and `||`) without needing Embers or another add-on to Handlebars in order to evaluate the arguments passed to the Handlerbars helper.

For non-programmers, the javascript looks different than templates, but the different paradigm is similar to average Excel users putting data in simple tables and being unfamiliar with SQL.  The javascript I wrote is not very complex since the result (each server config) is not that complex; evaluate some conditions and construct a string of server the configuration.  However, the javascript is not a template and does not look like a template.  To see the end result, the code should be run.

Experiment created for discussion:
https://github.com/gstrauss/ssl-config-generator/tree/js-helpers/src/templates/partials
https://github.com/gstrauss/ssl-config-generator/tree/js-helpers/src/js

(While I dabble in Javascript, it is not my primary language, so I am open to others sharing better idioms and approaches.)

src/templates/partials/lighttpd.hbs
gstrauss: Things will be simpler if ssl-config-generator stopped "recommending" TLS1 and TLSv1.1 (for the Old config) since TLSv1.2 or better is STRONGLY recommended and often required in modern professional settings.

janbrasna: That has nothing to with the fact that instead of simply transforming the output object to the config result, this tries to sniff into its content and adjust based on the literal value … for what I would perceive as no real gain. Compared to elsewhere:

https://github.com/mozilla/ssl-config-generator/blob/66667fa52ae1990ab289c9fe49b53fa5b3cf1422/src/templates/partials/postgresql.hbs#L18-L20

🤷

janbrasna: @gstrauss Your main concern is forward compatibility by leaving out the defaults. In this case you generate a config for a version N, but just leaving out the defaults at that time, for version N, unknown what may come later. If that config content gets carried over to version N+1, can you still assure it won't break the expectations and recommendation tiers? (=You can't.) So I see the value in being explicit by not having the reality change behind the scenes for the user, as consistency and repeatability is one of the benefits of these configs.

If you configured `intermediate` some time/versions ago (with TLS1.2 min left out as implicit), you'd basically get it disabled and be forced to `modern` unknowingly over time and lighttpd updates once TLS1.2 is not the min default anymore. And that's not a feature, that's definitely a bug — the config won't be stable over time. (And you're aware of  that, it's not an omission but definitely intentional… and I'm not willing to introduce intentional breakage or time bombs just to appease authors' perceptions of their own configs or defaults chosen — which are sort of irrelevant to the output of this tool.)

_(This is borderline unreviewable for me as i can't even be sure the logic is not flawed… spending so much extra time on something i don't think should ever be the scope of the tool here…)_

My opinion is this is doing the consumers disservice, but I don't see any value spending any more time trying to review it responsibly when I see it won't make any difference, so just the usual compromise:

Same as elsewhere, if you want to insist on not setting the value, at least emit it commented out with the note it's the default mentioning version numbers too — so when that changes over time, folks still know what is/was/should have been the explicit value and/or can uncomment it if there's a mismatch in their expectations.
(=Have them yell at server authors for changing things around. Not yell at @mozilla for generating crappy config boilerplates.)

gene1wood (dismissed):

src/templates/partials/lighttpd.hbs
```diff
@@ -38,19 +16,58 @@ ssl.privkey = "/path/to/private_key"
...
+  {{/if}}
   {{#if output.ciphers.length}}
+   {{#if (minver "1.4.68" form.serverVersion)}}
+    {{#if (includes "old" form.config)}}
+ssl.openssl.ssl-conf-cmd += ("CipherString" => "{{{join output.ciphers ":"}}}")
+    {{else}}
+
+# lighttpd TLS defaults are widely supported by clients and should be preferred
+# See https://wiki.lighttpd.net/Docs_SSL
+# Uncomment to better match the less restricted Mozilla {{form.config}} spec.
```
janbrasna, line 58: Wording +cipher naming compatibility +deeplink to PFS docs:

```suggestion
# lighttpd uses secure TLS defaults, see https://wiki.lighttpd.net/Docs_SSL#Perfect-Forward-Secrecy-PFS
# If you need to match Mozilla {{form.config}} config exactly, uncomment below (OpenSSL compatibility only)
```

src/templates/partials/lighttpd.hbs
janbrasna: To demonstrate what I mean by inconsistent results:

I generate intermediate config for current 1.4.76 and use that — I get intermediate.
I use the same config in 1.4.78 once it comes out — and surprise, I get modern.
I have to come back to generate intermediate again this time for the newer 1.4.78 — and only then get intermediate back.

Basically the claim is the configs are more future proof, but in reality it makes them really compatible only with the version they were generated for — as there's so much logic that tries hard to leverage version-by-version changes to default, that once generated, it really should not be used with different server version, as you get different results than it was generated for.

I don't see it as positive, but at least with the comments stating moz config defaults at time of output and hardcoding the version it was generated for it should not confuse people too much.

src/templates/partials/lighttpd.hbs
gstrauss: > lighttpd uses secure TLS defaults

The lighttpd TLS defaults encompass more than simply the cipher list, so linking to PFS section is too specific for the intention of the statement.

> If you need to match Mozilla {{form.config}} config exactly, uncomment below (OpenSSL compatibility only)

"OpenSSL compatibility only" is true of ssl-config-generator.  lighttpd supports more than a handful of TLS libraries.  While the suggestion is truthful, it is terse and likely to be confusing to consumers.  I think it an overspecification of unnecessary information.  If you use it with a different lighttpd TLS module instead of lighttpd mod_openssl, then it will either work or it will not work.  In the latter case, lighttpd will emit an error, and the end-user can figure things out from there.  Alternatively, we could overspecify and point to the server-side-tls mapping of cipher names between various libraries, but I think that is better in the wiki and not exploded into generated configs.

janbrasna: if it's an argument why the ciphers are commented out, a link to the default ciphers list is a convenience. it's unnecessary to just let people figure that out on their own where that info is.

notes for openssl applicability was iirc taken from similar place elsewhere, where rules were only provided for that implementation.

src/templates/partials/lighttpd.hbs
janbrasna: NB: This changeset already includes the future defaults changing in a version yet unreleased — actually demonstrating the effect of the concerns above right here in the outputs how an `intermediate` today will just stop working for TLSv1.2-only clients once you cross the version numbers hardcoded in the template to emit different configs, diverting from the guidelines in favor of server defaults on purpose again — but if you come back here and punch in your new version number, you'll have this time bomb removed and have a matching config again just fine. Escapes me.

We should try to widen the compatibility with different versions as much as we're able to, not making it as narrow and specific as possible. 🤷

I don't see the value in forcing people to come back to re-generate the template here after every upgrade.

janbrasna:
> It's certainly possible that you could restructure things better than it was before but I can confidently say that using Handlebars greatly simplified the design

I don't think we realistically would move away from using templates to turn the data into various forms of outputs (the charm of templating is the basic format is the target conf/yaml/ini/toml, with just the tokens adding some decisions here and there) — my guess is it's just the limits of Handlebars logic that started this, but the answer is not rewrite all of that into some bespoke spaghetti string manipulation, but address the individual limitations — probably considering a move from Handlebars to Mozilla's own [Nunjucks](https://mozilla.github.io/nunjucks/templating.html#macro) (pretty much the same format, minimal impact, but adding the ability to set macros locally and call them with different values multiple times etc., set new variables, refer super() blocks etc. like in jinja2 that's used heavily around the org's projects), or the more low–level [EJS](https://ejs.co/#docs) that we already use in this project FWIW for the main render (that basically blends the JS with what's outside of the tokens, not adding any logic, you just write any javascript… but that has some migration overhead if we don't tweak the syntax options etc.)

> > Some other src/templates/partials/*.hbs templates use {{! ... }}. Have you tried that? Also, Handlebars has ~ modifier on the directive after the {{ to affect whitespace of template directives. 

I know the syntax;) the issue is when you start combining comments with whitespace control — most IDEs and syntax highlighters in general won't recognize the comments as comments then, which makes it a [crappy](https://github.com/janbrasna/ssl-config-generator/commit/93d9c84) experience. I removed most of my comments for that reason.

> I rewrote all the .hbs partials into javascript in a development branch. The result saves over 70 KB (!) size in javascript over the Handlebars-generated javascript and also allows some shared code as well as simplification of conditions since javascript supports multiple conditions (e.g. `&&` and `||`) without needing Embers or another add-on to Handlebars in order to evaluate the arguments passed to the Handlerbars helper.

I'm sorry I don't follow what's the relation to EmberJS that we don't have here, and such framework use won't improve any of that (yes, any async approach and callbacks make the helpers a pain to use, and need decoupling from the webpack loader to point all that to own runtime instance manually etc. but we don't have to worry about that here). For performance there's template precompilation that I'm not sure we use. Macros and more logic is a HBS design choice that was OK when the templates started, but over time it turned out we'd need more; NJK or EJS covers that. In the meantime I quickly [added](https://github.com/janbrasna/ssl-config-generator/commit/569144c) a few helpers that I needed, and have no issues [passing](https://github.com/janbrasna/ssl-config-generator/commit/7d7dd20) another helpers' output to them. 🤷 

> For non-programmers, the javascript looks different than templates, but the different paradigm

There's a lot of effort put into templating solutions to separate any output boilerplate from the actual code. I don't believe we should be storing the configuration formats with heavy escaping inside a waterfall of string operations.

src/templates/partials/lighttpd.hbs
gstrauss: >  (if you had a better heading to link, I'd happily do so, but it's all tucked away under PFS)

The lighttpd TLS docs will be updated for the Jan 2025 release which changes TLS defaults to `MinProtocol => "TLSv1.3"`.  An improper deep link might be broken, or might not longer point to the relevant portion of the document.

> The note for OpenSSL–only compatibility was adapted from the same line of the existing config ... only simplified for brevity (comment length) to fit with the rest of the instructions

That was not clear to me, and I am intimately familiar with the material, so your "simplification for brevity" was not helpful.

gstrauss:
> There's a lot of effort put into templating solutions to separate any output boilerplate from the actual code. I don't believe we should be storing the configuration formats with heavy escaping inside a waterfall of string operations.

There is a balance to be struck.  Templates make sense when the majority of the text is reused and the logic to fill in the template is relatively simple.  Programmatic string concatenation makes sense where most of the output is conditional strings or when the logic to fill in the template might be more complex.

gstrauss:
@janbrasna I recognize that you disagree with and do not see the value in providing simpler guidance to people consuming ssl-config-generator.  I do see the value and have quite a bit of experience supporting configuration questions for lighttpd.  We can disagree on opinions here.

I am going to move forwards with this PR.  However, things are not set in stone.  Future adjustments can be made based on feedback from users.

gstrauss:
> > Some other src/templates/partials/*.hbs templates use {{! ... }}. Have you tried that? Also, Handlebars has ~ modifier on the directive after the {{ to affect whitespace of template directives. Still, using a real programming language like javascript gives explicit control of the whitespace, instead of having to fight with the templating framework Handlebars, which is built on top of javascript.
> 
> Just wanted to pop in to say that the previous SSL Configuration Generator, circa 2015 or so, _did_ use Javascript to generate its configurations. It was so complicated and messy that it was basically unusable after only supporting Apache and nginx and nobody wanted to touch it.
> 
> It's certainly possible that you could restructure things better than it was before but I can confidently say that using Handlebars greatly simplified the design of the configuration generator and is what let it support as many things as it does today.

@april is there some place I could see the .js-based SSL Configuration Generator, circa 2015?  This git repo has history only back to your initial commit in 2019, which is tagged v4.0 in the repo.  If you have a moment, I'd also be interested in your high-level opinion of the *part* of SSL Config Generator that I converted to javascript underneath Handlebars:

> Experiment created for discussion:
> https://github.com/gstrauss/ssl-config-generator/tree/js-helpers/src/templates/partials
> https://github.com/gstrauss/ssl-config-generator/tree/js-helpers/src/js
> 
> (While I dabble in Javascript, it is not my primary language, so I am open to others sharing better idioms and approaches.)

Most of the .js helpers that I created from `template/partials/*.hbs` took me about 10-15 mins each, so I do not think it as onerous to maintain as what you described from 2015: "It was so complicated and messy that it was basically unusable after only supporting Apache and nginx and nobody wanted to touch it."

PR #197, opened 2023-04-10
Add support for Kubelet

atombrella:
I need to work a bit with this, but it's a good first take. Those settings are what the [CIS Kubernetes Benchmarks](https://www.cisecurity.org/benchmark/kubernetes) recommend.


docs/index.html
```diff
@@ -186,7 +186,14 @@ <h5 class="d-none d-md-block">&nbsp;</h5>
...
 
-          
+            <div class="form-check">
+              <input class="form-check-input" type="radio" name="server" id="server-kubelet" value="kubelet" >
+              <label class="form-check-label" for="server-kubelet">
+                Kubelet/Kubernetes
+              </label>
+            </div>
+
+
```
gene1wood, line 196: For changes to the `index.html` you'll want to make them in `templates/index.ejs` as `index.html` is rendered from the template.

src/templates/partials/kubelet.hbs
```diff
@@ -0,0 +1,14 @@
...
+tlsCertFile: /path/to/ca.crt
+tlsPrivateKeyFile: /path/to/privkey.pem
+tlsCipherSuites: '{{{join output.ciphers ","}}}'
+{{#if (includes "intermediate" form.config)}}
+tlsMinVersion: TLSv1.2
+{{else if (includes "modern" form.config)}}
+tlsMinVersion: TLSv1.3
+{{else}}
+tlsMinVersion: TLSv1.0
+{{/if}}
```
gene1wood, line 14: Instead of checking the configuration type like this, check the `output.protocols` variable like you can see here

https://github.com/mozilla/ssl-config-generator/blob/c78c9462a5fbc96dad77ca92defb742092277c9c/src/templates/partials/apache.hbs#L43-L45

gene1wood (changes requested):

docs/index.html
janbrasna: @atombrella This will be generated from configs.js values

src/templates/partials/kubelet.hbs
janbrasna: @atombrella An example with min- value might be e. g.:

https://github.com/mozilla/ssl-config-generator/blob/36671610cd85a132b7c425a9ba2ed0f7057f92fd/src/templates/partials/dovecot.hbs#L19

janbrasna (changes requested):
@atombrella Have you tried this output in a live config, without errors, and with the constraints really working (i. e. verifying via nmap, curl -vvv --tls* or ssllabs et al. that the intended values are really getting picked up correctly)? Just from glancing the output configs I see some bits that wouldn't work, probably leaving the values unconfigured.

_(From the PR description I get it's just a starting point, not sure if you're ready to review… If you need to mark it as a WIP for now you can change it to a draft PR…)_

src/js/configs.js
```diff
@@ -194,6 +194,16 @@ tomcat: {
     tls13: '8.0.0',
     usesOpenssl: false,
   },
+  kubelet: {
+    cipherFormat: 'go',
+    highlighter: 'yaml',
+    tls13: '1.14',
```
janbrasna, line 200: I can only reference TLS 1.3 support since v1.19 — could you please provide the resource to confirm v1.14 is correct?

src/templates/partials/kubelet.hbs
```diff
@@ -0,0 +1,8 @@
+# {{output.header}}
+# {{{output.link}}}
+apiVersion: kubelet.config.k8s.io/v1beta1
+kind: KubeletConfiguration
+tlsCertFile: /path/to/ca.crt
+tlsPrivateKeyFile: /path/to/privkey.pem
+tlsCipherSuites: '{{{join output.ciphers ","}}}'
+tlsMinVersion: {{output.protocols.[0]}}
```
janbrasna, line 8: The protocols need to be converted into Go format, see: https://github.com/mozilla/ssl-config-generator/pull/197#discussion_r1453940826

janbrasna:
```suggestion
tlsMinVersion: {{#if (eq output.protocols.[0] "TLSv1")}}VersionTLS10{{else}}{{{replace output.protocols.[0] "TLSv1." "VersionTLS1"}}}{{/if}}
```

src/js/configs.js
janbrasna: @atombrella +please order the items alphabetically — tls13 comes before usesOpenssl…

janbrasna (changes requested):
@atombrella Do you need any more detailed help with the outstanding review bits to be addressed?

TODO:
- [x] TLSv1.3 support https://github.com/mozilla/ssl-config-generator/pull/197#discussion_r1453952519
- [ ] correct TLSv1.3 version? https://github.com/mozilla/ssl-config-generator/pull/197#discussion_r1482345879
- [x] ciphers as a sequence https://github.com/mozilla/ssl-config-generator/pull/197#discussion_r1453978984
- [ ] test ciphers array length https://github.com/mozilla/ssl-config-generator/pull/197#discussion_r1453935188
- [ ] protocols need to use golang format https://github.com/mozilla/ssl-config-generator/pull/197#discussion_r1482346951
- [ ] json items ordering https://github.com/mozilla/ssl-config-generator/pull/197#discussion_r1482353207
- [ ] kubelet config ordering https://github.com/mozilla/ssl-config-generator/pull/197#discussion_r1482631484

src/templates/partials/kubelet.hbs
```diff
@@ -0,0 +1,8 @@
+# {{output.header}}
+# {{{output.link}}}
+apiVersion: kubelet.config.k8s.io/v1beta1
+kind: KubeletConfiguration
+tlsCertFile: /path/to/ca.crt
+tlsPrivateKeyFile: /path/to/privkey.pem
+tlsCipherSuites: '{{{join output.ciphers ","}}}'
```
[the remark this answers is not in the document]
janbrasna:
```suggestion
{{#if output.ciphers.length}}tlsCipherSuites: [{{{join output.ciphers ","}}}]{{/if}}
```

src/js/configs.js
janbrasna: (actually you have to place the whole `kubelet: { … }` entry in its correct position alphabetically, otherwise it would get rendered between Tomcat and Traefik…)

janbrasna: You can see the current build of the PR here: https://mje-k8s--mozsslconf-dev.netlify.app/#server=kubelet to play around with it and see the results.

atombrella:
@janbrasna @gene1wood Thank you for the feedback!

I'm sorry it took me a while to get back to this. I've posted a couple of screenshots from my local setup. The SSL configurator does indeed start, and you  can select either Docker has been helpful 
to avoid cluttering my local setup; I left a PR to add a Dockerfile to the repository, but haven't received any comments/review.
I did not know about the netlify functionality.

The configuration file doesn't seem to be very common. Instead, the kube-apiserver.yaml is more common. Thus, I've posted the recommendation that kube-bench 
gives. The tool implements all of the guidelines from the CIS Kubernetes Benchmark.

```
4.2.13 If using a Kubelet config file, edit the file to set `TLSCipherSuites` to
TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_GCM_SHA256
or to a subset of these values.
If using executable arguments, edit the kubelet service file
/etc/systemd/system/kubelet.service.d/10-kubeadm.conf on each worker node and
set the --tls-cipher-suites parameter as follows, or to a subset of these values.
--tls-cipher-suites=TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_GCM_SHA256
Based on your system, restart the kubelet service. For example:
systemctl daemon-reload
systemctl restart kubelet.service
```

One question that is a bit vague to me, is whether I should leave out the cipher suite parameter for TLS 1.3, instead
of keeping it blank.

janbrasna:
@atombrella I've already written the suggested array check above in https://github.com/mozilla/ssl-config-generator/pull/197#discussion_r1482358380 for you incl. the brackets. Same with golang's TLS version formats https://github.com/mozilla/ssl-config-generator/pull/197#discussion_r1482351808, if that makes sense you can click-to-commit directly from the suggestion diff.

(Netlify is currently not linked to this repo but I'm slowly working on it. The preview is from my QA pull testing…)

_(The dockerfile PR needs also some love so that's not priority right now; since recently the stack has been updated to hopefully enable higher node versions to plug into automation, so when I have some time I'll check node 20 or LTS compatibility, that's higher priority to work on local machines with newer engines than 12 or 14…)_

src/js/configs.js
```diff
@@ -194,6 +194,15 @@ tomcat: {
...
   },
+  kubelet: {
+    cipherFormat: 'go',
+    highlighter: 'yaml',
+    hasVersions: false,
+    name: 'Kubernetes',
+    supportsHsts: false,
+    supportsOcspStapling: false,
+    usesOpenssl: false,
+  },
```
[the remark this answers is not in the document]
atombrella: That has done.

gstrauss:
Added Kubelet link to this issue to:
https://github.com/mozilla/ssl-config-generator?tab=readme-ov-file#requested-but-not-yet-added-new-software-support

atombrella:
Good. I don't really have the bandwidth/energy to finish this. I think Kubernetes from Azure, DigitalOcean, AWS and GCP all have this implemented correctly.

PR #200, opened 2023-05-08
Fix incorrect href in 5.0 JSON guideline

gene1wood:


gene1wood:
I fixed this mistake in the wrong location. The `/docs` directory is rendered  from the `/src` directory. I should have made this change in `/src` and then rendered a new version with npm in the `/docs` directory

PR #203, opened 2023-05-10
Add Coturn configuration

JGoutin:
Add [Coturn](https://github.com/coturn/coturn) (TURN and STUN Server) configuration.

"Modern" tested on Fedora 36 with Coturn 4.6.2.
"Intermediate" tested on Fedora 36 to 38 with Coturn from 4.5.2 to 4.6.2 (Used this config for a while).


src/js/configs.js
```diff
@@ -43,6 +43,15 @@ module.exports = {
...
     usesOpenssl: false,
   },
+  coturn: {
+    highlighter: 'ini',
+    latestVersion: '4.6.2',
+    name: 'Coturn',
+    showSupports: false,
+    supportsHsts: false,
+    supportsOcspStapling: false,
+    tls13: '3.2.2',
```
gene1wood, line 53: It looks like TLS 1.3 support was added via https://github.com/coturn/coturn/pull/989 

I don't see it mentioned in https://github.com/coturn/coturn/blob/master/ChangeLog before this point. Can you point me to where you found that TLS 1.3 was supported in coturn 3.2.2?

JGoutin: I don't find anymore where I found this. So it's safest to use the version that publish https://github.com/coturn/coturn/pull/989 : 4.6.2

gene1wood:
@JGoutin Thanks for the contribution! I appreciate it.

gene1wood:
This is deployed

PR #204, opened 2023-05-15
Add TLS_DHE_RSA_WITH_CHACHA20_POLY1305_SHA256 cipher to intermediate configuration

gene1wood:
This adds the `TLS_DHE_RSA_WITH_CHACHA20_POLY1305_SHA256` / `DHE-RSA-CHACHA20-POLY1305` cipher to the end of the intermediate cipher lists for openssl and iana.

You can see the specific cipher list change in b2a5a068fbfbe46007c16dd6528c7ada56bcf20f

See related PR https://github.com/mozilla/server-side-tls/pull/291

Fixes mozilla/server-side-tls#285


martinthomson (approved):
I checked this out and there is a versioned symlink in the `src/static/guidelines` directory that probably needs to be updated as well.  Approval is conditional on fixing that up.

```
lrwxrwxrwx 1 martin martin     8 May 16 11:31 latest.json -> 5.6.json
```

That's all I can see (you could update the changelog to point to this commit, but then that might not work out once this merges in).

gene1wood:
> I checked this out and there is a versioned symlink in the src/static/guidelines directory that probably needs to be updated as well. Approval is conditional on fixing that up.

Good catch, thanks for pointing that out. Fixed in f842855f970e23ffeb030029d6cffc90ab953dc8

> That's all I can see (you could update the changelog to point to this commit, but then that might not work out once this merges in).

Indeed, since we don't have versioned tags or releases in this repo (I figured that I didn't want to start versioning the entire repo just for the guideline) I figured I'd go in after and update the changelog to point to the right commit.

Long term, we may want to version the guideline outside of this repo so it can have tagged version, distinct from the version of the ssl-config-generator.

PR #214, opened 2023-12-26
Fix broken favicons

janbrasna:
Adding just a single png favicon borrowed from protocol favicon using the most simple format in just the only available size to link to something instead of 404s.

Back in the day there was an ico borrowed from wiki linked when server-side-tls was published to gh-pages, but that linking has been long lost. No assets have been provided since the split despite being defined in the template.


gene1wood:
Thanks @janbrasna !

PR #218, opened 2024-01-03
Update minimist to 1.2.8

gene1wood:
The current version of handlebars (4.7.8) depends on minimist but specifies the vulnerable version (1.2.5).

Until handlebars 5 comes out, we need to set a minimist version manually. After it's out we can remove the dependency on minimist

https://github.com/handlebars-lang/handlebars.js/issues/1851

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,7 @@
         "handlebars": "^4.7.8",
         "highlight.js": "^10.7.3",
         "jquery": "^3.7.1",
+        "minimist": "^1.2.8",
         "popper.js": "^1.16.1",
         "semver": "^6.3.1"
       },
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "handlebars": "^4.7.8",
     "highlight.js": "^10.7.3",
     "jquery": "^3.7.1",
+    "minimist": "^1.2.8",
     "popper.js": "^1.16.1",
     "semver": "^6.3.1"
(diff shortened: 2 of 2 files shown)


janbrasna:
@gene1wood Instead of adding it manually you could just run npm audit to fix the version only in package-lock and keep it bumped there.

gene1wood:
@janbrasna sounds good to me, I'll do that, thank you.

gene1wood:
@janbrasna Actually I can't seem to get `npm audit fix` to identify and fix the vulnerability in `minimist` 1.2.5 for some reason. 

(There is a tangential issue that we're using `browser-sync` `3.0.2` which doesn't work with the newest version of `browser-sync-webpack-plugin` `2.3.0` because https://github.com/Va1/browser-sync-webpack-plugin/pull/96 isn't merged and released yet, but even when rolling back to `browser-sync` `2.29.3` npm audit doesn't call out and offer to fix the minimist issue.

I'll just leave it like this (in an imperfect state) for the time being.

janbrasna:
Gotcha. I was about to test the browsersync v2-v3 compatibility myself too;) The minimist update should be easy to lock to a specific version [as dependabot tried](https://github.com/mozilla/ssl-config-generator/pull/193/files) but who knows — handlebars maintainers state that any clean npm install should pull the fixed version correctly, no idea why an audit fix doesn't resolve it.

PR #225, opened 2024-01-04
Update Apache stapling version

janbrasna:
I haven't found any reference why restrict to 2.4.14+ — bugs were in 2.4.10 and 2.4.11 implementations, good performance is reported since 2.4.13; this updates it accordingly.


gene1wood:
Looks good. Reference here : https://cwiki.apache.org/confluence/display/httpd/OCSPStapling#OCSPStapling-FixesrelatedtoOCSPStaplingindifferent2.4.xlevels

PR #228, opened 2024-01-05
Join output protocols

janbrasna:
Replaces variety of `each/unless` historic constructs with more readable `join()` helper that was added later where nothing more complex is necessary (mysql, nginx, proftpd, redis, tomcat) mainly to get rid of fighting the separators and wrestling with negative space tricks. (More complex iterations for XML or quotes were left intact.)

_(Tested, no change to output.)_


gene1wood:
Looks great, thanks.

PR #250, opened 2024-10-04
Add matrix strategy to test node versions building PRs

janbrasna:
Until preview builds are added here, this at least shows some basic feasibility of package updates and nodejs version compatibility on the webpack build output.

Screen Shot 2024-10-04 at 23 04 01

`test-build #1` workflow run e.g. [step:5:350](https://github.com/mozilla/ssl-config-generator/actions/runs/11186792442/job/31102543824#step:5:350)

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -1,0 +1,25 @@
+name: test-build
+
+on:
+  workflow_dispatch: # allow manual runs
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [18.x, 20.x, 22.x]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


gstrauss:
FYI: https://endoflife.date/nodejs notes that security support for nodejs 18.x ends 30 Apr 2025

gstrauss (approved):

janbrasna:
> support for nodejs 18.x ends 30 Apr 2025

At which point Node 23 should be out and somewhat battle-tested to add it instead of the EOLed v18. This nonetheless only tests basic functionality for production builds (mainly to report early if any incompatibility with the automation runners should bite us at some point in the future) — the dev/watch stack might raise compatibility issues for contributors separately: the actual stable env for local work might be slightly more conservative (but documented nonetheless!) than the supported runner images and their cached node versions used in deployment workflows.

(NB: I still have set v14.x for the Netlify preview build env from back in the day, as a quick sanity check to see when we lose compatibility with the lower teens…)

Issue #238, opened 2024-04-09
Postfix 3.7+ custom DH params discouraged on OpenSSL 3


wolegis:
See https://www.postfix.org/postconf.5.html section **smtpd_tls_dh1024_param_file**

> With Postfix ≥ 3.7, built with OpenSSL version is 3.0.0 or later, if the parameter value is either empty or "auto", then the DH parameter selection is delegated to the OpenSSL library, which selects appropriate parameters based on the TLS handshake. This choice is likely to be the most interoperable with SMTP clients using various TLS libraries, and custom local parameters are no longer recommended when using Postfix ≥ 3.7 built against OpenSSL 3.0.0.

gstrauss:
Included in #256 is a change for Postfix 3.7+ and OpenSSL 3.0.0+

janbrasna:
> "custom local parameters are no longer recommended"

Well, I believe they mean random arbitrary params, those are def not recommended — but I see no mention of that option being deprecated or anything alike — it's still perfectly fine to use it and provide the safe primes as we do now.

I'd much rather control the groups than leave that to RFC 7919 negotiation allowing #162 bad performance:/

EDIT: Ah, okay, sane defaults here, luckily: https://www.postfix.org/postconf.5.html

> **tls_ffdhe_auto_groups** (default: see "postconf -d" output)
> 
> The prioritized list of finite-field Diffie-Hellman ephemeral (FFDHE) key exchange groups supported by the Postfix SMTP client and server. OpenSSL 3.0 adds support for FFDHE key agreement in TLS 1.3. In OpenSSL 1.1.1, TLS 1.3 was only supported with elliptic-curve based key agreement. The "tls_ffdhe_auto_groups" parameter makes it possible to configure the list of FFDHE groups that the Postfix client or server will enable in OpenSSL 3.0 and up. This parameter has no effect when Postfix is built against earlier OpenSSL versions.
> 
> The **default list of FFDHE groups that Postfix enables in OpenSSL 3.0 and up includes just the 2048 and 3072-bit groups**. Stronger FFDHE groups perform poorly and EC groups are a much better choice for the same security level. Postfix ignores group names that are unknown to OpenSSL, or that are known but not yet implemented. The FFDHE groups are largely a backup, in case some peer does not support EC key exchange, or EC key exchange needs to be disabled for some pressing reason.

[...]

PR #252, opened 2024-10-06
Allow nginx ≥1.23.2 `ssl_session_tickets`

### LINKED ISSUE #69: Nginx ssl_session_tickets with modern on v5
As `modern` is now only `TLSv1.3` perhaps it doesn't make a sense for that to even specify:
```
ssl_session_tickets off;
```

As that is not used with `TLSv1.3`.

Yes, I know it is part of `generic` section, so please close if not considered with any importance.

### LINKED ISSUE #239: nginx http2 in listen-Directive is obsolete
Hello,

you generate for modern nginx:

```
# generated 2024-04-09, Mozilla Guideline v5.7, nginx 1.17.7, OpenSSL 1.1.1k, modern configuration, no HSTS, no OCSP
# https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=modern&openssl=1.1.1k&hsts=false&ocsp=false&guideline=5.7
server {
    listen 443 ssl http2;
    listen [::]:443 ssl http2;
    ...}
```

This is obsolete.
Should be:

```
server {
    listen 443 ssl;
    listen [::]:443 ssl;
    http2 on;
    ...}
```

See https://nginx.org/en/docs/http/ngx_http_v2_module.html

janbrasna:
Disabled in https://github.com/mozilla/server-side-tls/pull/80 (as explained in [archive](https://wiki.mozilla.org/Security/Archive/Server_Side_TLS_4.0#TLS_tickets_.28RFC_5077.29) and in detail https://github.com/mozilla/server-side-tls/issues/135), this can now be left out for nginx:

> "TLS session tickets encryption keys are now automatically rotated when using shared memory in the `ssl_session_cache` directive."

— _https://trac.nginx.org/nginx/milestone/nginx-1.23.2 (https://nginx.org/en/CHANGES-1.24)_

Fixes https://github.com/mozilla/server-side-tls/issues/284, https://github.com/mozilla/server-side-tls/issues/282, and tangentially resolves #69
(Also fixes #239 along the way by updating the versions used…)

### DIFF
--- a/src/js/configs.js
+++ b/src/js/configs.js
@@ -113,12 +113,12 @@ module.exports = {
   nginx: {
     checked: true,
     highlighter: 'nginx',
-    latestVersion: '1.17.7',
+    latestVersion: '1.26.0',
     name: 'nginx',
     tls13: '1.13.0',
   },
   openssl: {
-    latestVersion: '1.1.1k',
+    latestVersion: '1.1.1w',
     tls13: '1.1.1',
   },
   oraclehttp: {
--- a/src/templates/partials/nginx.hbs
+++ b/src/templates/partials/nginx.hbs
@@ -30,11 +30,13 @@ server {
     ssl_certificate_key /path/to/private_key;
     ssl_session_timeout 1d;
     ssl_session_cache shared:MozSSL:10m;  # about 40000 sessions
+{{#unless (minver "1.23.2" form.serverVersion)}}
 {{#if (minver "1.0.2l" form.opensslVersion)}}
   {{#if (minver "1.5.9" form.serverVersion)}}
     ssl_session_tickets off;
   {{/if}}
 {{/if}}
+{{/unless}}
 
 {{#if output.usesDhe}}
     # {{output.dhCommand}} > /path/to/dhparam


src/js/configs.js
```diff
@@ -113,12 +113,12 @@ module.exports = {
   nginx: {
     checked: true,
     highlighter: 'nginx',
-    latestVersion: '1.17.7',
+    latestVersion: '1.26.0',
     name: 'nginx',
     tls13: '1.13.0',
   },
   openssl: {
-    latestVersion: '1.1.1k',
```
gstrauss, line 121: Latest openssl is 3.3.2.  The OpenSSL Software Foundation Inc currently supports openssl 3.x and later.  1.1.1 and earlier are EOL from the OpenSSL Software Foundation Inc (though Premium Support (paid) may still be available for 1.1.1w).  https://openssl-library.org/post/2023-09-11-eol-111/

Generating config for openssl < 3.0 might be something to consider for an "older" snapshot of the site, but most popular modern distros all ship openssl 3.x and so we should be producing configs for currently support versions of openssl.

gstrauss: I am going to resolve this here since it is an openssl comment, and not an nginx comment.  There should be a separate issue opened for updating the openssl latest version to 3.3.2 after a review that existing directives recommended for apps using openssl are compatible with openssl 3.x.

gstrauss (approved):

src/js/configs.js
janbrasna: There are unfortunately known issues with OpenSSL 3 compatibility, so we're far from any official support at the moment.

gstrauss: Could you point me at any documented compatibility issues with respect to ssl-config-generator?  I am initimately familiar with openssl 3.x interface changes and deprecations, and as I mentioned above, most *modern* popular distros ship the supported openssl 3.x instead of earlier versions.

janbrasna: Yes, mainly not fiddling with dhparam and leaving that to the lib to infer the correct size. That however needs the server to either support that, or allow configuring the context to set that to auto. Which depends case by case, so may need gathering a bit of feedback for a while. (Servers relying on the new api might throw deprecation warnings for using the 1.1.1 route etc., therefore before any label reads "3.x.x" as the supported version, the configs must actually be 3+ ready at that time;)… hence why I kept it at the 1.1.1 branch, where no breaking changes in OP value defaults interfering with the applicability of the configs etc. happen — not to make any false promises for now.)

The other issue would be explaining how to support `old` config if one wishes, as TLS1–1.1 now need SECLEVEL=0:
> "The security strength of SHA1 and MD5 based signatures in TLS has been reduced. This results in SSL 3, TLS 1.0, TLS 1.1 and DTLS 1.0 no longer working at the default security level of 1 and instead requires security level 0."

I should probably open a meta/tracking issue for that actually, to keep it in one place.

gstrauss: We should probably open two issues: one for openssl, and the other for following up with mozilla to refresh the specification recommendations.  For ssl-config-generator to remain useful it must support openssl 3.x.  Otherwise, will become an historical curiosity or merely a reference for outdated, less secure systems.

gstrauss: #238 already quotes the suggestion to avoid specifying dhparams with openssl 3.

https://openssl-library.org/post/2024-09-10-tls-deprecation/ reminds us that the security level must be reduced to 0 for TLS1.0 and TLS1.1, and that support will also (likely) need to be explicitly selected at build time for openssl 3.5.

PR #254, opened 2024-10-09
Fix deploy-to-production artifact action version

janbrasna:
Fixup #236

### DIFF
--- a/.github/workflows/deploy-to-production.yml
+++ b/.github/workflows/deploy-to-production.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Build with Webpack
         run: npm run build
       - name: Upload artifact
-        uses: actions/upload-pages-artifact@v4
+        uses: actions/upload-pages-artifact@v3
         with:
           path: ./docs
 


gstrauss (approved):

gstrauss:
@janbrasna FYI, you may assign yourself as a reviewer and then self-approve the review.  Then you'll be able to merge.

gstrauss:
Now I see updates reflected in https://ssl-config.mozilla.org/

I hadn't caught the upload-pages-artifact version when reviewing.  Thanks for fixing it!

gstrauss:
Sorry, I should have caught the failure in the github actions error log.  Thanks for fixing it!

janbrasna:
@gstrauss Well, NOW I can;) but the protections were there before, even with 0 required approvals the default branch was still protected from pushing PRs that the actor is the last to push to it, irrelevant who is the author or reviewer (and authors were not allowed approving reviews, either…) — but generally I'd like to only self-r+ on things I broke myself to just fix up, or if something needs shipping like right now for a reason.

janbrasna:
Actions validation is hard, some things are caught by the editor, some not, and some validations only trigger after pushing to master, so… eh. The same will be with Dependabot configs, they usually take a few attempts to actually get right;) so no worries, this needs a few dry runs off of a fork's default branch to get the real annotations et al., and this was a stupid last-minute addition that seemed like a no brainer. Well…

PR #256, opened 2024-10-10
Lower SECLEVEL in `old` config for OpenSSL 3.x compatibility

### LINKED ISSUE #188: Using OpenSSL 3.0.1 and TLS 1.1 and lower will usually fail
Support for TLS 1.1 and lower was moved to security level 0 in [OpenSSL 3.0.1](https://www.openssl.org/news/openssl-3.0-notes.html#:~:text=SSL%203%2C%20TLS%201.0%2C%20TLS%201.1%2C%20and%20DTLS%201.0%20only%20work%20at%20security%20level%200%2C%20except%20when%20RSA%20key%20exchange%20without%20SHA1%20is%20used.). This implies by default usually those protocolls will not be supported without further configuration.

## Reproduce

* Server Software: nginx
* Mozilla Configuration: Old
* Server Version: 1.23.0
* OpenSSL Version: 3.0.1

The configuration will generate a nginx configuration that lists TLS 1.1 and 1.0, but those protocols will usually not work when OpenSSL 3.0.1 or higher is used at the same time.

```
    ssl_protocols TLSv1 TLSv1.1 TLSv1.2 TLSv1.3;
```

Starting an nginx server with this configuration and issuing a TLS 1.1 or TLS 1.0 request against it will result in a SSL error like the following:

```
2023/01/02 17:41:13 [info] 39#39: *2 SSL_do_handshake() failed (SSL: error:0A000076:SSL routines::no suitable signature algorithm) while SSL handshaking, client: 192.168.192.3, server: 0.0.0.0:443
```

Related: https://github.com/nginxinc/docker-nginx/issues/743

gstrauss:
Updates to support OpenSSL 3.x

Resolves #188
Tracked in #260

* Set SECLEVEL=0 in cipherstring for `old` output to allow TLSv1–v1.1 for OpenSSL 3.x
* ~Hide DH parameters for servers calling OpenSSL interfaces `SSL_CTX_set_dh_auto()` or `SSL_set_dh_auto()` to use RFC7919.~ (TLS1.2 will keep just ffdhe2048, TLS1.3 will have to limit curves/groups, followup TBD)

- - - - -

TESTING:
https://test-3x--mozsslconf-dev.netlify.app/#openssl=3.3.3&config=old&server=postgresql

### DIFF
--- a/src/js/state.js
+++ b/src/js/state.js
@@ -41,6 +41,11 @@ export default async function () {
   } else {
     ciphers = ciphers;
   }
+  if (ciphers.length && ciphers[0] === '@SECLEVEL=0') ciphers.shift();
+  if (configs[server].usesOpenssl !== false && minver('3.0.0', form['openssl'].value)) {
+    // set SECLEVEL=0 via cipher string to support TLSv1-1.1 "old" with OpenSSL 3.x
+    if (protocols.includes('TLSv1.1')) ciphers.unshift('@SECLEVEL=0');
+  }
 
   const state = {
     form: {


gstrauss:
For DH parameters code review, I checked all servers which had .hbs configs containing `{{#if output.usesDhe}}`.  I downloaded the source code and checked the history to find supported versions.

lighttpd - no change; does the right things with the appropriate versions

apache - small changes to do the right things with appropriate versions
postfix - small changes to do the right things with appropriate versions
redis - small changes to do the right things with appropriate versions

coturn - no change; does not call SSL_CTX_set_dh_auto/SSL_set_dh_auto
dovecot - no change; does not call SSL_CTX_set_dh_auto/SSL_set_dh_auto
exim - no change; does not call SSL_CTX_set_dh_auto/SSL_set_dh_auto
haproxy - no change; does not call SSL_CTX_set_dh_auto/SSL_set_dh_auto
nginx - no change; does not call SSL_CTX_set_dh_auto/SSL_set_dh_auto
postgresql - no change; does not call SSL_CTX_set_dh_auto/SSL_set_dh_auto
proftpd - no change; does not call SSL_CTX_set_dh_auto/SSL_set_dh_auto
squid - no change; does not call SSL_CTX_set_dh_auto/SSL_set_dh_auto

There are some apps, such as stunnel and postgresql, which use built-in 2048-bit pregenerated DH params, if none is otherwise configured.  Some other servers many support DH params included in certificate PEM file.  This PR does not review bit sizes of DH params as that is out of scope for this PR -- which is focused on upgrading the OpenSSL version supported by ssl-config-generator.  Besides, best practices in modern configs no longer use DHE.

gstrauss:
RFE: I would like to discussion renaming "Old" config to "Insecure", because that is what it is.  It provides obscurity over clear-text, but is not considered "secure" by modern specifications.  It fails all modern security standards and so the remaining value is compatibility with ancient devices and possibly historical verification of now-obsolete older security standards.

For TLSv1 and TLSv1.1 (also no longer considered secure), @SECLEVEL=0 in the cipher list should suffice.  Another alternative would be suggesting modifications to `/etc/ssl/openssl.cnf` or `/etc/pki/tls/openssl.cnf`, but I think modifying the cipher string scopes SECLEVEL=0 to the app, which is better than suggesting lowering the OpenSSL library security for all apps on the system using openssl.

janbrasna:
I will be addressing this gradually, mostly looking into some template options to avoid the repetition, TBD…

> I would like to discussion renaming "Old" config to "Insecure", because that is what it is. It provides obscurity over clear-text, but is not considered "secure" by modern specifications. 

That name is a term within the org:
https://infosec.mozilla.org/guidelines/risk/scoring_and_other_levels.html#recommended-configuration-states

A lot of effort went (back in the day) to make this config usable, not having anything directly compromised etc., so it's still a nudge above "insecure" in the scoring (yellow, grade "D" to "D-"); basically anything moved out of `old` in the past decade can be deemed insecure (red, grade "F") in comparison. I believe the [rationale](https://wiki.mozilla.org/Security/Server_Side_TLS#Old_backward_compatibility) below the list still stands. (The issue is none of that is directly visible in the UI here, that is definitely a deficiency.)

Given the specs are due for update anyways, I can see some `old` cleanup and demotions from `intermediate` in the cards eventually… so no need to re-label it with catastrophic badges, it only needs to move the suites along gradually.

> For TLSv1 and TLSv1.1 (also no longer considered secure), @SECLEVEL=0 in the cipher list should suffice. 

There will be some servers that won't know how to support that (individual keys for ciphers instead of the cipher string; will need to look into the implementations), but happy to have old configs on 3+ only as a best effort, if documented as such… 

Do we also want to map when a server doesn't support 3+? (Does it still happen today?)

> This PR does not review bit sizes of DH params as that is out of scope for this PR

I'm actually comfortable with letting intermediate fail to apply on <80 bits — that's a reminder for the consumer they're not careful with the key size as instructed.

src/js/state.js
```diff
@@ -41,6 +41,14 @@ export default async function () {
   } else {
     ciphers = ciphers;
   }
+  if (form['config'].value === 'old') {
+    if (configs[server].usesOpenssl !== false) {
+      // set SECLEVEL=0 for openssl 3 to support TLSv1 and TLSv1.1 for Old config
+      if (minver('3.0.0', form['openssl'].value) === true) {
+        ciphers.unshift('@SECLEVEL=0');
```
janbrasna, line 48: Haven't looked into properly, only quick observation: https://gst-os3--mozsslconf-dev.netlify.app/#server=oraclehttp&version=12.2.1&config=old&openssl=3.3.2&guideline=5.7

```nginx
# old configuration
SSLProtocol             All 
SSLCipherSuite          @SECLEVEL=0:@SECLEVEL=0:@SECLEVEL=0:@SECLEVEL=0:@SECLEVEL=0:@SECLEVEL=0:@SECLEVEL=0:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-
```

Not that it wouldn't work, right? (I haven't tried tbh;D) but it just doesn't look right.

src/js/state.js
```diff
@@ -41,6 +41,14 @@ export default async function () {
   } else {
     ciphers = ciphers;
   }
+  if (form['config'].value === 'old') {
```
janbrasna, line 44: Could that be inferred from `protocols` perhaps at this point already, instead of hardcoding it? (in case the profiles change over time…)

src/js/state.js
gstrauss: That is funny.  I believe I had tested with stunnel.hbs, and hadn't looked across the rest.   I'll see about modifying the javascript tomorrow or later in the weekend.

src/js/state.js
gstrauss: I think it would be more fragile for the code to infer it, since then this code may need to be modified when the ssl-config-generator spec is (eventually) updated.  By contrast, if we were to rename profiles, that is easier to identify profile name use in the code.  I came across the config name use in caddy.hbs earlier today, so there is at least that also referencing the profile name.

gstrauss:
In the future, if we do consider renaming the profiles, I would like to propose including a date in the spec.  Even if only a year, "modern-2020" is much more descriptive than the current almost 5-year old (!) "modern", which I do not consider modern.

> > For TLSv1 and TLSv1.1 (also no longer considered secure), @SECLEVEL=0 in the cipher list should suffice.
> 
> There will be some servers that won't know how to support that (individual keys for ciphers instead of the cipher string; will need to look into the implementations), but happy to have old configs on 3+ only as a best effort, if documented as such…

Most of the app templates join the cipher list with ':'.  The ones that do not are awselb caddy go jetty traefik, so we'll have to look to see how they handle (or not) `@SECLEVEL=0`.  I was trying to avoid creating a new variable for @SECLEVEL=0 that had to be pasted onto the cipher list and have code added to each and every `src/templates/partials/*.hbs` for addition to the cipher lists (where applicable).

> 
> Do we also want to map when a server doesn't support 3+? (Does it still happen today?)

gstrauss:
> > This PR does not review bit sizes of DH params as that is out of scope for this PR
> 
> I'm actually comfortable with letting intermediate fail to apply on <80 bits — that's a reminder for the consumer they're not careful with the key size as instructed.

The default `SECLEVEL=2`, and we should retain that.  At `SECLEVEL=1`, TLSv1 and TLSv1.1 are disabled, so if the Old config needs them, those who need it must use `SECLEVEL=0`

src/js/state.js
gstrauss: I pushed a change to fix this.  If you're concerned with the condition `if (ciphers[0] !== '@SECLEVEL=0') {` on a possibly empty ciphers list (which should not happen for the Old config), then I can change the logic to something similar used to set `usesDhe` further below in the same routine: `if (ciphers.join(":").includes("@SECLEVEL=0") === false) {`.

gstrauss: Not an urgent concern of mine, but the repetition that I now have to check for suggests that there are some performance gains on the client-side to be had by eliminating unnecessary repetition of work.  Obviously, things need to be re-evaluated, when a config change is made by the user, but it should not need to be re-evaluated 5 times.

gstrauss:
To address your review comment about apps that do not pass a ':'-separated cipher string to openssl, I pushed a change to avoid adding `@SECLEVEL=0` to the cipher list for `['awselb','caddy','go','jetty','traefik']`.  We'll need to add a comment in each of this .hbs files that `@SECLEVEL=0` will need to be configured in `openssl.cnf` unless there is a different way that these apps expose to configure the security level in their config files.  (not done yet)

gstrauss:
['awselb','caddy','go','jetty','traefik'] all are configured `usesOpenssl: false`, so I removed that unnecessary check.

I also reconsidered and change the `form['config'].value === 'old'` test to `protocols.includes('TLSv1.1') === true` since `SECLEVEL=0` is needed by OpenSSL 3 for TLSv1 and TLSv1.1 used by the Old configuration.

gstrauss:
@janbrasna I think this is ready for commit.  Please give this priority for re-review.  Thanks!

I feel it would be negligent to unnecessarily delay recommending configs supporting OpenSSL 3.x.  On the other hand, if there are are few bugs that turn up in specific apps for Old configuration -- which we want to discourage anyway due to low security -- then those are bugs and can be fixed.

gstrauss:
I incorporated the changes in #257 and updated the Apache version here.

gstrauss:
Ready for review again with the changes from #257.  The last few pushes were whitespace cleanup.

gstrauss:
> I'm not sure about Tomcat impact

No impact.  Tomcat is configured in src/js/config.js with `usesOpenssl: false` and that is checked before adding `SECLEVEL=0`

>> "This may have more implications than just re-enabling TLSv1.0 and 1.1 ...

> Can you think of any wild differences among versions or distros/defaults?

From a practical perspective, it does not matter since it is required to support TLSv1 and TLSv1.1, and if the Old config provides support for TLSv1 and TLSv1.1, then `SECLEVEL=0` must be set if OpenSSL 3.x is used.

> I would love for this to be first released as "opt-in" — say by skipping the 3.x version in configs.js now

Ok.  I will push momentarily with that commit removed.

janbrasna:
> > > This PR does not review bit sizes of DH params as that is out of scope for this PR
> > 
> > I'm actually comfortable with letting intermediate fail to apply on <80 bits — that's a reminder for the consumer they're not careful with the key size as instructed.
> 
> The default `SECLEVEL=2`, and we should retain that. At `SECLEVEL=1`, TLSv1 and TLSv1.1 are disabled, so if the Old config needs them, those who need it must use `SECLEVEL=0`

Ah right, I meant <112 bits, sorry. That kinda enforces the RSA and DHE keys length 2048b+ for us, so I'm not sure there's anything to review as that's on par with the recommendations (and good it's actually enforced now) — did you mean anything beyond that?

gstrauss:
My earlier comment was addressing the poster's recommendation to use at least ffdhe3072.  That is not in scope for this PR and will be addressed in the future with specification review. :)

src/js/state.js
janbrasna: I was basically more concerned about _how_ it got iterated over so many times in the first place;)

> I was trying to avoid creating a new variable for @SECLEVEL=0 that had to be pasted onto the cipher list and have code added to each and every hbs

Sure, that is similar to tls13 logic also filtering out stuff in state.js before passing it on opaquely to the template output. That's definitely a good place!

(BTW, it's plain JS here, not HBS, so you don't have to nest the conditions so bizarrely (:) and can use falsy bool values without too much remorse really…)

src/js/state.js
```diff
@@ -41,6 +41,14 @@ export default async function () {
...
     ciphers = ciphers;
   }
+  if (configs[server].usesOpenssl !== false && minver('3.0.0', form['openssl'].value) === true) {
+    // set SECLEVEL=0 for openssl 3 to support TLSv1 and TLSv1.1 for Old config
+    if (protocols.includes('TLSv1.1')) {
+      if (ciphers[0] !== '@SECLEVEL=0') {
+        ciphers.unshift('@SECLEVEL=0');
+      }
+    }
+  }
```
janbrasna, line 51: is the condition broken down for linting (line/length) reasons? or a simple:

```suggestion
  if (configs[server].usesOpenssl && protocols.includes('TLSv1.1') && minver('3.0.0', form['openssl'].value)) {
    // set SECLEVEL:0 via cipher string to support TLSv1-1.1 "old" with OpenSSL 3.x
    if (!ciphers.includes('@SECLEVEL=0')) ciphers.unshift('@SECLEVEL=0');
  }
```

could do?

src/js/state.js
janbrasna: Quite the opposite, once the protocols fall off the cliff, this will stop lowering the seclevel for us (I'd prefer to test on features… profile names should only be used for decorative purposes, they don't bear any logic per se… caddy example is one of the worst sins;D)

Or you mean there are other algos that may need the 80–112bit range even once the weak protocols are out? That's a fair point, but TBH something we can't assess right now, when we don't know the next "old" specs, and what feature might need the seclevel tuning (and then, I'd still like to test for such feature if possible, instead of the whole profile… let's see what the next json update brings.)

src/js/state.js
janbrasna: Well, empty ciphers may happen if there's no overlap of the recommendation with the ciphers available from the server.  So maybe even:
```suggestion
  if (configs[server].usesOpenssl && protocols.includes('TLSv1.1') && minver('3.0.0', form['openssl'].value) && ciphers.length) {
    // set SECLEVEL:0 via cipherstring to support TLSv1-1.1 "old" with OpenSSL 3.x
    if (!ciphers.includes('@SECLEVEL=0')) ciphers.unshift('@SECLEVEL=0');
  }
```
?

src/templates/partials/redis.hbs
```diff
@@ -14,8 +14,15 @@ tls-ca-cert-file /path/to/ca_certificates.crt
...
 {{#if output.usesDhe}}
+{{#if (minver "6.2.7" form.serverVersion)}}
+{{#unless (minver "3.0.0" form.opensslVersion)}}
 # {{output.dhCommand}} > /path/to/dhparam
 tls-dh-params-file /path/to/dhparam
+{{/unless}}
+{{else}}
+# {{output.dhCommand}} > /path/to/dhparam
+tls-dh-params-file /path/to/dhparam
+{{/if}}
```
janbrasna, line 25: OK would you be willing to add an "AND" helper (e.g. https://github.com/helpers/handlebars-helpers/blob/master/lib/comparison.js …) to avoid this repetition all over the place, please?

src/js/state.js
gstrauss: I am not a javascript developer, so I appreciate any recommendations for better coding idioms or coding standard preferences.

Please add to your wishlist an issue to ask why this code is being run repeatedly for some templates.  It did not appear to run repeatedly for the (new) stunnel template.

src/js/state.js
gstrauss: yes, testing for feature in updates to this PR

src/js/state.js
gstrauss: > is the condition broken down for linting (line/length) reasons?

Yes.

> Well, empty ciphers may happen if there's no overlap of the recommendation with the ciphers available from the server.

Not with the current "Old" config.  I expect any update to the "Old" config which still includes TLSv1 or TLSv1.1 will include a non-empty ciphers list.

In any case, I incorporated some of your changes, but still avoided excessively long lines.

gstrauss:
> verify exim doesn't need removal of +no_ssl* flags

I verified the exim code.

> decide on tomcat (that may or may not use openssl implementation…)

src/js/configs.js specifies `usesOpenssl: false`

gstrauss:
> review all use of output.usesDhe, output.dhCommand, output.dhParamSize

I did review this.  Please be more specific for what you/I might look.

src/templates/partials/redis.hbs
gstrauss: Seems doable.  Time allowing, I'll try to revisit this on Sun.

If this existing code in this PR works correctly, I would prefer this PR not be blocked for this kind of polishing.

gstrauss:
I wrote:
> My earlier comment was addressing the poster's recommendation to use at least ffdhe3072. That is not in scope for this PR and will be addressed in the future with specification review. :)

I should note that some servers hard-code ffdhe2048.  stunnel is one.  I think postgresql was another.  (I was not keeping track as I quickly reviewed source code for calls to `SSL_CTX_set_dh_auto()` or `SSL_set_dh_auto()`.)

lighttpd is a third, but lighttpd only uses ffdhe2048 if openssl version is too old.  The reason I have not updated lighttpd is because if you're on an ancient system and using an ancient openssl version, you are probably on an underpowered system and you likely have many other system security risks more important than DHE bit length.  If this is a concern, you should, of course, configure lighttpd with your own appropriate `DHParameters`.

gstrauss:
A large amount of time is being spent on `DHParameters`, even though modern security recommendations do not include ciphers which require them.  I do not think we would be spending time on this if the Mozilla TLS recommendations were reviewed and updated.

Note:  https://developers.cloudflare.com/ssl/edge-certificates/additional-options/cipher-suites/recommendations/ recommends ECDHE and does not recommend any DHE ciphers, even for Legacy.  (If TLSv1 and TLSv1.1 are needed, then weaker ciphers are made available for such Legacy use.)

gstrauss:
> add an AND helper for conditional blocks only (not) matching two values

That is an underlying code change which has not effect on the output.  It should not be part of this PR and should be tested and rolled out separately, in a separate PR.

> decide on tomcat (that may or may not use openssl implementation…)

As already stated, this is not an issue with the current code which sets `usesOpenssl: false`.  If you want to change that for Tomcat, that IS NOT PART OF THIS PR and should not block this PR.

> verify oraclehttp support

https://docs.oracle.com/middleware/12213/webtier/administer-ohs/GUID-C76BCA2A-9C28-4D16-9758-9346FBCF7512.htm#HSADM1016

I don't have access to test oraclehttp, but if someone reports an issue with the Old config and SECLEVEL=0, then we can add code to oraclehttp.hbs to filter out SECLEVEL=0.  Please remember that this is for the Old config; the Old config is old and people are strongly discouraged from using it.

---

@janbrasna Please focus on the "minimum viable product" to approve this PR so that it can be deployed for testing.  This PR *does not* bump the openssl version, so you'll continue to have plenty of time for turd polishing for the Old config after this PR is merged and deployed.  Please move as many of your testing concerns as you can from this PR to #263

polarathene:
> recommendation to use at least ffdhe3072

2048-bit is still plenty, using the hashrate of the bitcoin global network in 2024 [I did some napkin math](https://www.reddit.com/r/selfhosted/comments/1f67n7q/comment/llal9lm/) (_see near end of comment for 2048-bit math_) which calculated 54,000 years, but that was just against the entropy with the hashrate as operations. I don't have math on me from old notes for GFNS which also requires considerable resources beyond CPU at this scale, but IIRC the attack cost should be higher... thus really not pragmatic / worthwhile.

TLS 1.3 also supports ffdhe2048 as a minimum IIRC and that's auto-negotiated? Usually you only see a desire to use more because the user does not understand the math, or some regulation compliance has raised the bar without adequate justification (_presumably as a precaution or external perception/expectation_).

src/templates/partials/redis.hbs
gstrauss: Turns out to not be trivial.
Examples in https://stackoverflow.com/questions/8853396/logical-operator-in-a-handlebars-js-if-conditional work on simple values but not on items which need to be evaluated.

Some possible (not tested) solutions include  https://github.com/jmurphyau/ember-truth-helpers and or `lvalue = Embers.Handlerbars.get(this, lvalue) || lvalue` or https://gist.github.com/servel333/21e1eedbd70db5a7cfff327526c72bc5

Possible solutions need to be evaluated for security XSS.

A better solution, IMHO, is to hoist logic which needs multiple conditions out of the templates and into javascript, and have the templates access the result.  I don't think adding more dependencies and bloating the generated page is the best solution.  Please create a separate github issue if you would like to evaluate options.  At the moment, the small bits of duplication in the templates is a workable solution for this PR.

gstrauss:
Rebased on master.

After discussion in meeting, I have removed the dh_auto changes from this PR. (Note, this also removed the update to Apache version).  Removing the dh_auto changes from this PR preserve the existing behavior for Intermediate to explicitly set 2048-bit DHE key.  TLSv1.3 uses RFC7919 negotiation to select Group, and whether or not and how to configure Groups (Curves) will be a separate discussion from this PR.  In the ssl-config-generator, `usesDhe` is not set for TLSv1.3.

The remaining commit in this PR sets `@SECLEVEL=0` for "Old" config to support TLSv1 and TLSv1.1 for OpenSSL 3.x.  (SECLEVEL=1 would be minimum needed for 1024-bit DHE key used by "Old" config.)

src/js/state.js
```diff
@@ -41,6 +41,12 @@ export default async function () {
   } else {
     ciphers = ciphers;
   }
+  if (configs[server].usesOpenssl && minver('3.0.0', form['openssl'].value)) {
+    // set SECLEVEL=0 via cipher string to support TLSv1-1.1 "old" with OpenSSL 3.x
+    if (protocols.includes('TLSv1.1')) {
+      if (!ciphers.includes('@SECLEVEL=0')) ciphers.unshift('@SECLEVEL=0');
+    }
+  }
```
janbrasna, line 49: Looking at line lengths around (as L34) with similar conditionals, there's no formatting enforced really, but the order of evaluation mostly preferred, so I went for speed and simplification and e.g. this one stops evaluating once the leftmost is not relevant, so the protocol should go first etc.:
 
```suggestion
  if (protocols.includes('TLSv1.1') && configs[server].usesOpenssl && minver('3.0.0', form['openssl'].value)) {
    // to allow "old" TLSv1-1.1 with OpenSSL 3.x we add SECLEVEL=0 via cipher string
    if (!ciphers.includes('@SECLEVEL=0') && ciphers.length) ciphers.unshift('@SECLEVEL=0');
  }
```

Also checking that we're actually adding it to an existing list, not creating a bogus list from empty result (there may not necessarily be an overlap of the guideline and the actual server able to offer own list of ciphers)

janbrasna: Ah, with the defaults not set in configs, the usesOpenssl value is not truthy, it's just unset, so needs precise testing… sorry.

```suggestion
  if (protocols.includes('TLSv1.1') && configs[server].usesOpenssl !== false && minver('3.0.0', form['openssl'].value)) {
    // to allow "old" TLSv1-1.1 with OpenSSL 3.x we add SECLEVEL=0 via cipher string
    if (!ciphers.includes('@SECLEVEL=0') && ciphers.length) ciphers.unshift('@SECLEVEL=0');
  }
```

src/js/state.js
```diff
@@ -41,6 +41,10 @@ export default async function () {
   } else {
     ciphers = ciphers;
   }
+  if (protocols.includes('TLSv1.1') && configs[server].usesOpenssl !== false && minver('3.0.0', form['openssl'].value)) {
+    // to allow "old" TLSv1-1.1 with OpenSSL 3.x we add SECLEVEL=0 via cipher string
+    if (!ciphers.includes('@SECLEVEL=0') && ciphers.length) ciphers.unshift('@SECLEVEL=0');
+  }
```
janbrasna, line 48: And shift when switching states:

```suggestion
  }
  if (protocols.includes('TLSv1.1') && configs[server].usesOpenssl !== false && minver('3.0.0', form['openssl'].value)) {
    // to allow "old" TLSv1-1.1 with OpenSSL 3.x we add SECLEVEL=0 via cipher string
    if (!ciphers.includes('@SECLEVEL=0') && ciphers.length) ciphers.unshift('@SECLEVEL=0');
  } else {
    if (ciphers[0] === '@SECLEVEL=0') ciphers.shift();
  }

```

src/js/state.js
gstrauss: > Ah, with the defaults not set in configs, the usesOpenssl value is not truthy, it's just unset, so needs precise testing… sorry.

Ok.  Then your changes should have been minimal changes to address that, not rewriting other parts (yet again).  Please try to be much, MUCH, more precise in making the minimum necessary changes.

I wrote the conditional logic in a specific order for a reason.  The check for TLSv1.1 in protocols is only necessary for openssl 3.x where we have to change the security level.  You checking for TLSv1.1 in protocols first "because you wanted to" is not the logical progression that I communicated in the way that I coded it or the intention of this PR, which is titled "Updates to support OpenSSL 3.x".  Your change to reorder the logic was unnecessary and poor code review hygiene, besides a waste of time for all involved.  Please try to get things done more expediently.

src/js/state.js
gstrauss: This should be done unconditionally before the logic which might add it, and also needs to check ciphers.length, since ciphers.length might be 0, e.g. with the current config for Modern with TLSv1.3.

gstrauss:
I have incorporated the two changes: 1) test `usesOpenssl !== false` and 2) `if (ciphers[0] === '@SECLEVEL=0') ciphers.shift();` unconditionally.

src/js/state.js
janbrasna: The review suggestion was not addressed so I went on to make the changes myself to finish testing and be able to release this.

If OpenSSL 3.x is gonna be the default, it will always be that case, so doesn't make sense to test first, the earliest to bail is to test for protocols, don't understand the rest of your concerns (whether they are real or just hypothetical, you'd have to be more specific, sorry.)

If you don't like my reviews, don't request them. Thanks. It was not in a working state i.e. not even tested.

src/js/state.js
janbrasna: - No need as it's either–or, no other case so if it's there it's tested and not added again; no need to remove beforehand, and only need to remove if it's there and shouldn't (not touching it otherwise on other state changes)
- I originally had `.at(0)` to not have worry about indexes/prototype etc.; but that doesn't have good browser support — need to check if that's something `babel` takes care of for us.
- Not sure if happy trailing `ciphers?.[0]` evaluates better.
- Basically no matter what type the var is, accessing the index even for empty array or a number, string or bool ends up falsy (no error is raised if that's what you're concerned about), the shift only happens when it's safe to do.

src/js/state.js
janbrasna: Reasoning for order was in the first comment of this thread above the suggestion. Not "because I wanted to".

Reasoning for not creating just "@SECLEVEL=0" cipherstring from an empty list due to possible `configs[server].supportedCiphers` intersection was described in the first post below the code suggestion.

Your rebase removing these, on the other hand, doesn't seem to be explained anywhere; so I'll just leave it to the others — as indeed, going back for the third time un-solving the comments at any cost really is a waste of everyone's time.

src/templates/partials/redis.hbs
janbrasna: Thanks for taking a stab at these… that unfortunately sounds awfully familiar from when we were adding some heavy hbs logic few years ago. I will consult what we ended up using and with how much success regarding the overall experience and flexibility…

(I don't follow the Ember note as we probably won't be adding a framework to facilitate any rendering when we have the current webpack stack with loaders etc. serving that right from the single bundled app…)

EDIT: This has solved itself over time as the repetition for defaulting to auto groups for various versions is not in the scope of this PR anymore.

EDIT2: I had quickly adapted some of our helpers ([e.g.](https://github.com/janbrasna/ssl-config-generator/compare/9e96f08...7d7dd2), +[demo](https://fix-awselb-dhe--mozsslconf-dev.netlify.app/#server=awselb)) and had no issues besides some inverted logic from what I was used to. Ugly, only way without moving away from HBS first, but best you can do and it seems to work for what we need. (This example should not ever be necessary again, as it's an ugly hack for a reason; normally you need just a nice `(and … … …)` plus its inversion. It only needs some more work to turn the expression `{{when}}` returning the outputs, needing `{{#if (when …)}}`, to block proper, allowing `{{#when …}}{{/when}}` blocks directly.)

src/js/state.js
janbrasna: This is not related to `old` config, but the intersection with servers' support in `configs[server].supportedCiphers` that may end up empty, there's no guardrail around none overlapping (as it's then checked in the outputs)

src/js/state.js
janbrasna: FWIW instead of shifting an already manipulated cipher list, it'd be better to find why it's not being populated fresh on every state change from `ssc.ciphers.openssl` (that it looks like it should, it just keeps referencing the already changed version from the past state(?))

janbrasna:
> > I'm not sure about Tomcat impact
> 
> No impact. Tomcat is configured in src/js/config.js with `usesOpenssl: false` and that is checked before adding `SECLEVEL=0`

Which doesn't change whether folks end up using OpenSSL or JSSE. I don't have enough working knowledge to infer how the connectors configure own `OPENSSL_CONF` or the context from its own jdk/jre `java.security` properties or `-Djdk.*` args at env/startup etc. — I'm more looking at this whether we need to start listing some caveats when we don't know if such profile can be supported fully (like we do in caddy comments)

> A large amount of time is being spent on `DHParameters`

Because it's the most sensitive setting in the current outputs, and of the added value features we provide translating the guidelines into the configs.

> even though modern security recommendations do not include ciphers which require them

Which is not true. All TLSv1.3 suites include them by default (even though they don't "advertise" the fact in the IANA naming, but it's defined directly in the RFC).

polarathene:
> We still reflect what is in the current recommendations, and will try to do so even for RFC7919 groups, if we end up restricting them in the configs.

Yeah, I [mentioned some reasons](https://github.com/mozilla/ssl-config-generator/pull/256#issuecomment-2409098698) for why someone might feel the need to use a higher key/group size but as per the napkin math, I really don't see it being necessary.

Given that TLS 1.3 will handle many connections these days and IIRC a client or server may restrict it's ability to only negotiate with a cipher using FFDHE 2048-bit, since TLS 1.3 doesn't allow you to exclude that (not since I last checked anyway?) it'd be accepted 🤔 

Since it's a static group for any such TLS 1.3 connection negotiated with it, someone with the resources could attack that for an advantage. At which point some policy goes into place to forbid/change it, but as per the math I linked earlier it's not very pragmatic to pull that off:

> _Besides... what if someone has the equivalent of that global bitcoin hashrate of 750 EH/sec against RSA-2048 (110-bit)???:_
> 
> ```
> # 110-bit (RSA 2048-bit) operations processed at a rate of 750EH/sec
> # divided by seconds in a thousand years (approx), provides thousand of years required
> ( ((2^110) / 750e18) / (60*60*24*365*1e3)) ) ~= 54.88
> ```
> 
> _54,000 years!!! Panic! The LetsEncrypt certificate with 90 days of validity is in danger! Quick bump it up to RSA 3072-bit!_
> 
> _Don't try to bring in quantum computing either. For similar reasons to why the math for RSA above isn't valid (GNFS), there are certain resource requirements that offset the advantages you see parroted around._

The link provides more context/info for that, but you need an awfully large amount of computation and advancement to bring that attack into the realm of possibility, as covered in [another part of my napkin math reference](https://www.reddit.com/r/selfhosted/comments/1f67n7q/comment/llaljpv/):

> _Also for the 750 EH/sec, we don't have the best data on power usage, but it's estimated at 170 TW hours of electricity per year. More than Ukraine outputs annually._
> _At just 10x the size of current network, the power output would be nearing that of India, 30x would surpass the USA and 200x the entire planets output._
> _To support 1000x the bitcoin network size requires producing more energy that you're going to get how?_

The power required is sufficient to boil all the oceans on the planet, it's not something you can do stealthily even if we could compute an attack within our lifetimes.

Unfortunately we have silly articles like [this 2024 Forbes one](https://www.forbes.com/councils/forbestechcouncil/2024/02/23/the-current-encryption-landscape-the-need-for-3072-bit-keys/) encouraging 3072-bit and citing quantum compute risks:

> _While these key sizes are currently deemed secure against brute force attacks by computers, they may not provide security against quantum computers._
> _For instance, a quantum computer employing Shor's algorithm might [potentially crack](https://www.technologyreview.com/2019/05/30/65724/how-a-quantum-computer-could-break-2048-bit-rsa-encryption-in-8-hours/) a 2048-bit RSA key within eight hours._

Which to anyone who knows better this is nonsense. We are a long way off having the necessary qubits (_the hardware you see marketed these days tends to be a lot lower in actual qubits since many are used for error correction_). While the compute time could go down, it doesn't magically mess with physics, the entropy is still there and the resource cost just shifts, time is only one resource to take into consideration.

> _NIST Recommendations: The National Institute of Standards and Technology (NIST) explicitly recommends the use of 3072-bit keys as part of their guidelines for quantum cryptography._
> _Following these guidelines ensures compliance with recognized standards and industry best practices._

RSA keys, not FFDHE group parameters but close enough compute wise IIRC 😅 

That's the sort of thing that'll make a push for bumping it up, regardless of the math. That increase is smaller than RSA 1024-bit to 2048-bit too (_NIST has their own document on equivalent symmetric security strength_).

gstrauss:
If someone is using "Old", then the priority is not "best security practices"; the priority is compatibility with out-of-date, end-of-life, and/or unmaintained devices (or all of the above).  TLSv1.2 was released in 2008, over 16 years ago.  TLSv1.0 and TLSv1.1 have been formally deprecrated since March 2021 in [RFC8996](https://datatracker.ietf.org/doc/html/rfc8996).   Hello!  As I write this, over 3 1/2 years (!) has passed since March 2021 was published.

Therefore, I am committing this change as proposed over two weeks ago.  This change is small, and setting `SECURITY_LEVEL=0` is required for OpenSSL 3.x to support TLSv1.0 and TLSv1.1 protocols.

Follow-ups for https://github.com/mozilla/ssl-config-generator/pull/256#issuecomment-2458504186 in #270 "Explicitly configure curves/groups from the guidelines"

oraclehttp and Tomcat are mark in ssl-config-generator `src/js/configs.js` with `usesOpenssl: false` and are out-of-scope for the changes propsed in this PR.

@janbrasna, the your additions to the TODO list are out-of-scope for this PR.  Please handle elsewhere:
> TODO/OQ:
> - [x] ~consider AND helper for conditional blocks only (not) matching two values~
> - [x] move oraclehttp from mod_ssl to mod_ossl
> - [ ] look into tomcat seclevel options (for both apr/openssl implementation and java policy?)
> - [ ] how does e.g. BoringSSL like that (should not matter when it doesn't support some weak ciphers at all?)

gstrauss:
libressl supports `@SECLEVEL`.  BoringSSL does not recognize `@SECLEVEL`.  While both try to provide some level of compatibility with common OpenSSL interfaces, neither are OpenSSL.  libressl has its own version numbers.  BoringSSL is Google-specific and the source code README.md very explicitly states: (https://boringssl.googlesource.com/boringssl)
```
BoringSSL is a fork of OpenSSL that is designed to meet Google's needs.

Although BoringSSL is an open source project, it is not intended for general
use, as OpenSSL is. We don't recommend that third parties depend upon it. Doing
so is likely to be frustrating because there are no guarantees of API or ABI
stability.
```

Issue #125, opened 2021-02-15
apache: Allow use of SSLOpenSSLConfCmd DHParameters if supported


tim-seoss:
Versions of Apache > 2.4.8 support:
`SSLOpenSSLConfCmd DHParameters "/path/to/dhparams"`

I think a useful enhancement would be to emit this (if supported) rather than the current:
```
# curl https://ssl-config.mozilla.org/ffdhe2048.txt >> /path/to/signed_cert_and_intermediate_certs_and_dhparams
#SSLCertificateFile      /path/to/signed_cert_and_intermediate_certs_and_dhparams
#SSLCertificateKeyFile   /path/to/private_key
```
Since that's a bit awkward for people to automate if they use automatic SSL cert updating (e.g. via letsencrypt).

gstrauss:
Applications using openssl should prefer `SSL_CTX_set_dh_auto()` and related functions, available since OpenSSL 1.1.0, instead of providing DHParameters.
https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_dh_auto.html

> If "auto" DH parameters are switched on then the parameters will be selected to be consistent with the size of the key associated with the server's certificate. If there is no certificate (e.g. for PSK ciphersuites), then it it will be consistent with the size of the negotiated symmetric cipher key.
> 
> Applications may supply their own DH parameters instead of using the built-in values. This approach is discouraged and applications should in preference use the built-in parameter support described above.

@tim-seoss you might want to open an issue for Apache2 to request that they use `SSL_CTX_set_dh_auto()` or `SSL_set_dh_auto()`

I think that is better than end-users attempting to provide DHParameters (unless end-users know what they are doing and really want to do so).

fdelapena:
Moreover, `SSLOpenSSLConfCmd DHParameters "/path/to/dhparams"` is broken when using recent OpenSSL versions (e.g. 3.x) so it just won't work when just using it. [Apache Users Mailing List thread](https://lists.apache.org/thread/vpqvq8mk9smn0wv749c1ngj09lqbt2wx) and [Qualys Community thread](https://success.qualys.com/discussions/s/question/0D52L00006qPpqqSAC/strange-behavior-dhparam-apache-2451).

By the way, [internet.nl](https://internet.nl/) considers ffdhe2048 phased out on their test and recommends at least ffdhe3072.

gstrauss:
[...]

PR #257, opened 2024-10-10
Update Apache version for appending dhparam to cert file

janbrasna:
Fixup according to `mod_ssl` docs:

> "Custom DH parameters and an EC curve name for ephemeral keys, can also be added to end of the first file configured using [`SSLCertificateFile`](https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslcertificatefile). This is supported in version **2.4.7 or later**."

### DIFF
--- a/src/js/configs.js
+++ b/src/js/configs.js
@@ -7,7 +7,7 @@ const noSupportedVersion = '10000.10000.10000';
 module.exports = {
   apache: {
     highlighter: 'apache',
-    latestVersion: '2.4.41',
+    latestVersion: '2.4.60',
     name: 'Apache',
     tls13: '2.4.36',
   },
--- a/src/templates/partials/apache.hbs
+++ b/src/templates/partials/apache.hbs
@@ -12,7 +12,7 @@
 {{/if}}
 <VirtualHost *:443>
     SSLEngine on
-{{#if (minver "2.4.8" form.serverVersion)}}
+{{#if (minver "2.4.7" form.serverVersion)}}
   {{#if output.usesDhe}}
 
     # {{output.dhCommand}} >> /path/to/signed_cert_and_intermediate_certs_and_dhparams


gstrauss:
@janbrasna I have some changes in #256 which address #125 and use `SSLOpenSSLConfCmd DHParameters "/path/to/dhparams"`, which is supported starting in Apache 2.4.8.  If you really want this change in this PR for Apache 2.4.7 before `SSLOpenSSLConfCmd DHParameters` for Apache 2.4.8 and later, then I'll incorporate it into #125.  In any case, from Apache 2.4.52 and later with OpenSSL 1.1.0 and later, #256 lets Apache configure OpenSSL with `SSL_CTX_set_dh_auto()` and `SSL_set_dh_auto()`

gstrauss:
(I did not bump the Apache version in my PR)

janbrasna:
I'm not sure I'm confident to add `SSLOpenSSLConfCmd DHParameters` due to issues mentioned https://github.com/mozilla/ssl-config-generator/issues/125#issuecomment-1382698707 (but that comes from individual reports, which as well might be misconfigurations — but before we know more, I wanted to go the opposite way — update the concatenation version, skip `DHParameters` entirely, and only make use of the auto params added in the recent versions with #256) — I'll review that particular bit there separately, too.

> (I did not bump the Apache version in my PR)

That may render making use of some of the new logic a bit cumbersome for those not punching in their current version in the UI, so it should eventually be done in one place or another.

gstrauss:
> I'm not sure I'm confident to add `SSLOpenSSLConfCmd DHParameters` due to issues mentioned [#125 (comment)](https://github.com/mozilla/ssl-config-generator/issues/125#issuecomment-1382698707) (but that comes from individual reports, which as well might be misconfigurations

With the configs I proposed in #125, the reported issue has no effect for Apache 2.4.52 and later, where I omit `SSLOpenSSLConfCmd DHParameters` and use Apache 2.4.52 and later built-in support for RFC7919.  Note: Apache 2.4.52 was released Dec 2021, almost 3 years ago.

> > (I did not bump the Apache version in my PR)
> 
> That may render making use of some of the new logic a bit cumbersome for those not punching in their current version in the UI, so it should eventually be done in one place or another.

Yes, of course.

janbrasna:
> With the configs I proposed in #125, the reported issue has no effect for Apache 2.4.52 and later

And I would prefer to not add the `SSLOpenSSLConfCmd DHParameters` for <2.4.52 with it, that's what I'm pointing out showing anecdotal evidence of compatibility issues. (Maybe limiting that only to OpenSSL 1.1.0–1.1.1 for those older httpd? Is that extra logic worth maintaining for posterity?)

- - - -

Please everyone think about how badly we want to support separate `DHParameters` in Apache 2.4.8–2.4.52 (a ~3 year old release) basically only benefiting these who want to rotate certs and can't be bothered to append the param file — and who don't run a newer release at the same time.

It was not broken, only inconvenient. I'm not willing to fix inconveniences that have a chance of not working properly for folks (even if there's no canonical reference for that, only mentions on a mailing list).

I'd vouch for skipping from appending the params after the certs directly to the auto/RFC feature, leaving out this `SSLOpenSSLConfCmd DHParameters` when there are issues reported issues.

janbrasna:
Unrelated to this PR. It can wait how we resolve #125 or if we include #125 in #256 or not — I will review it there.

gstrauss:
Yann Ylavic is an Apache developer and wrote in
https://lists.apache.org/thread/vpqvq8mk9smn0wv749c1ngj09lqbt2wx
> Step 3) does not work anymore with latest openssl versions, the only
way to configure custom dhparams in httpd is to append them to the
certificate file

so I'll adjust #256 to not use `SSLOpenSSLConfCmd DHParameters` with OpenSSL 3.x

janbrasna:
I was not able to verify it elsewhere though (as in, what version, is there a tracking issue etc.?) — only referencing this mailing list post as the sole mention to avoid setting it on v3.x — but I generally trust such authors, somewhat unconditionally;) hence why I raised this to avoid it at all.

What would hypothetically make it worth is still adding that only for v1.1.0–1.1.1 — but, I'd vote to avoid that for the sake of template maintainability now, as it was not there the last few years when it was helpful — and not now.

(When it's decided `SSLOpenSSLConfCmd DHParameters` is not going in, this PR only fixes the formal version when the concatenation started working, nothing serious, only cosmetic change to match the docs — as there is no rationale to explain why we chose version different that the start of official support as stated by the authors.)

gstrauss:
I think the difference between the posts is that having dhparams in the CertificateFile was supported in Apache 2.4.7, and `SSLOpenSSLConfCmd DHParameters` in Apache 2.4.8.  Apache 2.4.7 was released in 2013, and Apache 2.4.8 was released in 2014.

gstrauss:
Merging this.  Scope of #256 was reduced.

PR #258, opened 2024-10-10
Update stunnel

### LINKED ISSUE #227: New stunnel template doesn't reflect selected configs
The newly added stunnel does not seem to work as intended:

> I'm not sure the outputs reflect chosen configuration. For the default render (`intermediate`, `latestVersion`) I seem to be getting `modern` no matter what config I select — unless I manually downgrade the server version below `tls13` when TLSv1.2-related output starts appearing reflecting parts of the chosen configuration level.
> 
> Is that as intended? It seems to hard-code the most restrictive flags based on the serverVersion and opensslVersion available, not taking the flexibility of the configs into account.
> 
> — _Originally posted by @janbrasna in https://github.com/mozilla/ssl-config-generator/pull/119#pullrequestreview-1805178201_

gstrauss:
Update stunnel

github: closes #227

### DIFF
--- a/src/templates/partials/stunnel.hbs
+++ b/src/templates/partials/stunnel.hbs
@@ -1,32 +1,34 @@
-# {{output.header}}
-# {{{output.link}}}
+; {{output.header}}
+; {{{output.link}}}
 
-{{#if (minver "1.0.1" form.opensslVersion)}}
-    {{#if (minver "5.50" form.serverVersion)}}
-        {{#if (minver "1.1.0" form.opensslVersion)}}
-            {{#if (minver "1.1.1" form.opensslVersion)}}
+{{#if (minver "5.50" form.serverVersion)}}
+{{#unless (includes "TLSv1" output.protocols)}}
+{{#unless (includes "TLSv1.1" output.protocols)}}
+{{#if (includes "TLSv1.2" output.protocols)}}
+sslVersionMin = TLSv1.2
+{{else if (includes "TLSv1.3" output.protocols)}}
 sslVersionMin = TLSv1.3
 ciphersuites = {{{join output.cipherSuites ":"}}}
-            {{ else }}
-sslVersionMin = TLSv1.2
-ciphers = {{{join output.ciphers ":"}}}
-            {{/if}}
-        {{ else }}
-options = NO_TLSv1
+{{/if}}
+{{/unless}}
+{{/unless}}
+{{else}}
+{{#unless (includes "TLSv1.1" output.protocols)}}
 options = NO_TLSv1.1
-ciphers = {{{join output.ciphers ":"}}}
-        {{/if}}
-    {{ else }}
+{{/unless}}
+{{#unless (includes "TLSv1" output.protocols)}}
 options = NO_TLSv1
-options = NO_TLSv1.1
-ciphers = {{{join output.ciphers ":"}}}
-    {{/if}}
-{{ else }}
-options = NO_SSLv2
+{{/unless}}
+{{/if}}
+{{#unless (minver "1.0.1" form.opensslVersion)}}
 options = NO_SSLv3
+options = NO_SSLv2
+{{/unless}}
+{{#if output.ciphers.length}}
+ciphers = {{{join output.ciphers ":"}}}
 {{/if}}
 
-; Using client mode to proxy gmail imap server
+; Example using client mode to proxy IMAP server
 [imap-proxy-server]
 client = yes
 accept = 127.0.0.1:143
@@ -37,6 +39,7 @@ checkHost = imap.example.com
 {{/if}}
 verifyChain = yes
 
+; Example serving HTTPS for HTTP service
 [https-server]
 accept  = 443
 connect = 80
--- a/src/js/configs.js
+++ b/src/js/configs.js
@@ -177,7 +177,7 @@ module.exports = {
   },
   stunnel: {
     highlighter: 'ini',
-    latestVersion: '5.71',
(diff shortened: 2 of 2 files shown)


gstrauss:
@gene1wood was the repo config changed recently?
> Review required
> New changes require approval from someone other than the last pusher. 

And yet, under the Reviewers sidebar on the right:
> At least 0 approving reviews are required to merge this pull request.

src/templates/partials/stunnel.hbs
```diff
@@ -1,29 +1,31 @@
 # {{output.header}}
 # {{{output.link}}}
```
janbrasna, line 2: INI syntax:

```suggestion
; {{output.header}}
; {{{output.link}}}
```

src/templates/partials/stunnel.hbs
```diff
@@ -1,29 +1,31 @@
...
+{{/if}}
+{{#unless (minver "1.0.1" form.opensslVersion)}}
 options = NO_SSLv3
+options = NO_SSLv2
+{{/unless}}
+{{#if output.ciphers.length}}
+ciphers = {{{join output.ciphers ":"}}}
 {{/if}}
 
 ; Using client mode to proxy gmail imap server
```
janbrasna, line 31:
```suggestion
; Example using client mode to proxy IMAP server
```

janbrasna: And if you don't mind perhaps add a comment before #L42 block too?

```ini
; Example serving HTTPS for HTTP service
[https-server]
```

janbrasna (commented):
Some preliminary improvements that fell thru QA when this was added:

gstrauss:
This merge fixes the issue in #227 where only the Modern config was displayed.

@gene1wood: it looks like the recent changes you made to the repo config have had the desired effect to remove the requirement for Approvals before merge.  Thanks!

janbrasna:
I would still prefer testing and approval from someone else than the author, but this was so broken it's better to publish now, and refine later. (I will be looking into formatting and also going through the version logic as described by the OP in their contribution — to follow what should be in the hbs and what's already taken care of in js — on a first glance this looks right, but will give it more than a first glance, however the priority is lower than the other things. I'd still generally expect a say ~14day window for reviews before self-approvals unless something's a hotfix though. Not a problem here, great job getting this sorted out so promptly, thanks!)

@gene1wood This deploy at least confirms the publishing source settings are A–OK and we're good to publish just the built artifacts now, and not compete with the static deploy;) Thanks!

src/templates/partials/stunnel.hbs
```diff
@@ -1,32 +1,34 @@
...
 
-{{#if (minver "1.0.1" form.opensslVersion)}}
-    {{#if (minver "5.50" form.serverVersion)}}
-        {{#if (minver "1.1.0" form.opensslVersion)}}
-            {{#if (minver "1.1.1" form.opensslVersion)}}
+{{#if (minver "5.50" form.serverVersion)}}
+{{#unless (includes "TLSv1" output.protocols)}}
+{{#unless (includes "TLSv1.1" output.protocols)}}
+{{#if (includes "TLSv1.2" output.protocols)}}
+sslVersionMin = TLSv1.2
```
janbrasna, line 8: We may need to explicitly set TLS1 as min version if present in `protocols` for 5.50+ though as its default may change anytime.

(Haven't checked the docs if they support it at all, what are the defaults etc. … but it may warrant a comment whether that version can be achieved at all if not etc.)

src/templates/partials/stunnel.hbs
```diff
@@ -1,32 +1,34 @@
...
+{{else if (includes "TLSv1.3" output.protocols)}}
 sslVersionMin = TLSv1.3
 ciphersuites = {{{join output.cipherSuites ":"}}}
-            {{ else }}
-sslVersionMin = TLSv1.2
-ciphers = {{{join output.ciphers ":"}}}
```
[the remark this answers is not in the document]
gstrauss: Please add to your wishlist to track.  I'm not going to consider these items until after the site is properly recommending security settings for contemporary software (e.g. openssl 3.x), not only EOL software versions (e.g. openssl < 3.x) and current versions of apps.

src/templates/partials/stunnel.hbs
gstrauss: Might consider `sslVersionMin = output.protocols[0]`
There are some *.hbs which use similar and assume the protocols list is sorted older to newer protocols, which is currently the case in src/static/guidelines/latest.json `tls_versions`.  TBD.

gstrauss:
> @gene1wood This deploy at least confirms the publishing source settings are A–OK and we're good to publish just the built artifacts now, and not compete with the static deploy;) Thanks!

@gene1wood would you like a separate issue opened and assigned to you to track these requests?
Please disable the static deploy (and document what was done in case it needs to be re-enabled).

janbrasna:
> > This deploy at least confirms the publishing source settings are A–OK and we're good to publish just the built artifacts
> 
> Please disable the static deploy

These were set correctly, and already working correctly, as can be confirmed from recent action runs. Thanks Gene, no action needed.

Screen Shot 2024-10-12 at 16 40 25

_(2/2 is correct, instead of the 5/5 before…)_

src/templates/partials/stunnel.hbs
janbrasna: also noticed the implicit reliance on array order: https://github.com/janbrasna/ssl-config-generator/issues/23

PR #261, opened 2024-10-10
Fix Caddy 2 configs, drop Caddy 1 support

### LINKED ISSUE #222: Updates needed for Caddy
@francislavoie made some good suggestions in #153 

I'm copying them here so we can consider them and potentially PR some changes based on them

@francislavoie writes

> Caddy v2 doesn't support lower than TLS 1.2 at all (because older TLS versions are completely broken). So all those clients won't work.
> 
> IMO the "old" option for Caddy should be totally disabled. The intermediate option should remove all tls config (irrelevant) and turning off TLS 1.2 with "modern" is kinda silly and counterproductive, so I'd also just remove the tls config for that option as well.
> 
> Caddy's defaults are secure, there's no reason to tune cipher suites, and configuring cipher suites has no effect at all when using TLS 1.3 because the Go stdlib automatically ordering them. See https://go.dev/blog/tls-cipher-suites as I mentioned earlier.
> 
> Also, Caddy doesn't use OpenSSL, the website makes it seem like it uses it by showing the OpenSSL version on the right. And Caddy v2.1.1 is a long-since EOL version.
> 
> Caddy v1 is no longer supported, so it does not make sense at all to continue showing config for it. Don't recommend config for EOL software, please.

gstrauss:
Update caddy

github: closes #222

### DIFF
--- a/src/templates/partials/caddy.hbs
+++ b/src/templates/partials/caddy.hbs
@@ -1,43 +1,31 @@
 # {{output.header}}
 # {{{output.link}}}
-{{#unless (includes "old" form.config)}}
+{{#unless (minver "2.0.0" form.serverVersion)}}
+# note that Caddy version 1 reached end-of-life > 4 years ago back in 2020
+{{else}}
+{{#if (includes "old" form.config)}}
+# note that Caddy supports only TLSv1.2 and later
+{{else}}
 # note that Caddy automatically configures safe TLS settings
+{{/if}}
 {{/unless}}
 
 # replace example.com with your domain name
-example.com
-{{! This is a big of a kludge due to Caddy restrictions on TLS versions and cipher suites }}
-{{#if (includes "old" form.config)}}
+example.com {
 
-tls {
-    protocols tls1.0 tls1.3
-    {{#if (minver "2.0.0" form.serverVersion)}}
+  tls {
+{{#if output.ciphers.length}}
+    # Due to a lack of DHE support, you -must- use an ECDSA cert to support IE 11 on Windows 7
     ciphers {{{join output.ciphers " "}}}
-    {{ else }}
-    ciphers ECDHE-ECDSA-AES128-GCM-SHA256 ECDHE-RSA-AES128-GCM-SHA256 ECDHE-ECDSA-AES256-GCM-SHA384 ECDHE-RSA-AES256-GCM-SHA384 ECDHE-ECDSA-WITH-CHACHA20-POLY1305 ECDHE-RSA-WITH-CHACHA20-POLY1305 ECDHE-ECDSA-AES128-CBC-SHA ECDHE-RSA-AES256-CBC-SHA ECDHE-RSA-AES128-CBC-SHA ECDHE-ECDSA-AES256-CBC-SHA RSA-AES128-CBC-SHA RSA-AES256-CBC-SHA RSA-3DES-EDE-CBC-SHA
-    {{/if}}
-}
-{{/if}}
-{{#if (includes "intermediate" form.config)}}
-
-# Due to a lack of DHE support, you -must- use an ECDSA cert to support IE 11 on Windows 7
-tls {
-    protocols tls1.2 tls1.3
-    {{#if (minver "2.0.0" form.serverVersion)}}
-    ciphers {{{join output.ciphers " "}}}
-    {{ else }}
-    ciphers ECDHE-ECDSA-AES128-GCM-SHA256 ECDHE-RSA-AES128-GCM-SHA256 ECDHE-ECDSA-AES256-GCM-SHA384 ECDHE-RSA-AES256-GCM-SHA384 ECDHE-ECDSA-WITH-CHACHA20-POLY1305 ECDHE-RSA-WITH-CHACHA20-POLY1305
-    {{/if}}
-}
 {{/if}}
 {{#if (includes "modern" form.config)}}
-
-tls {
     protocols tls1.3
-}
 {{/if}}
+  }
 {{#if form.hsts}}
 
-# HSTS ({{output.hstsMaxAge}} seconds)
-header Strict-Transport-Security "max-age={{output.hstsMaxAge}}"
+  # HSTS ({{output.hstsMaxAge}} seconds)
+  header Strict-Transport-Security "max-age={{output.hstsMaxAge}}"
 {{/if}}
+
+}
--- a/src/js/configs.js
+++ b/src/js/configs.js
@@ -37,7 +37,7 @@ module.exports = {
   caddy: {
     cipherFormat: 'caddy',
     highlighter: 'nginx', // TODO: find better
-    latestVersion: '2.1.1',
+    latestVersion: '2.8.4',
     name: 'Caddy',
     supportsOcspStapling: false, // actually true; can't be disabled in Caddy
     tls13: '0.11.5',


src/templates/partials/caddy.hbs
```diff
@@ -1,43 +1,31 @@
...
     ciphers {{{join output.ciphers " "}}}
-    {{ else }}
-    ciphers ECDHE-ECDSA-AES128-GCM-SHA256 ECDHE-RSA-AES128-GCM-SHA256 ECDHE-ECDSA-AES256-GCM-SHA384 ECDHE-RSA-AES256-GCM-SHA384 ECDHE-ECDSA-WITH-CHACHA20-POLY1305 ECDHE-RSA-WITH-CHACHA20-POLY1305
-    {{/if}}
-}
 {{/if}}
 {{#if (includes "modern" form.config)}}
-
-tls {
     protocols tls1.3
```
francislavoie, line 22: TLS 1.2 is still considered secure, especially with Caddy's default cipher list. Is it really worth suggesting to break compat with many clients with this option? Caddy's default are already "modern".

janbrasna: The naming does not come from "general hardening" or similar, the tool only reflects what is published in https://wiki.mozilla.org/Security/Server_Side_TLS

francislavoie: Fair enough :+1:

gstrauss: TLSv1.3 is widely supported by modern clients.  Unless sites must support older clients, new configurations can choose TLSv1.3-only, and as @janbrasna pointed out, these are the recommendations for the https://wiki.mozilla.org/Security/Server_Side_TLS "Modern" config.

In January, I plan to release lighttpd with the lighttpd default to use MinVersion TLSv1.3 unless otherwise configured.  This behavior change has been announced on lighttpd releases throughout 2024. 

However, I'll also note that cloudflare "Modern" recommendation still recommends both TLSv1.2 and TLSv1.3
https://developers.cloudflare.com/ssl/edge-certificates/additional-options/cipher-suites/recommendations/

gstrauss: We might consider including links to the specific "Modern", "Intermediate", "Old" specifications (including version) in the generated output.

gstrauss: I share your concern between HTTP servers and MTAs and other clients.  Instead of differentiating by types of services, we might have a "Modern" and a "Modern-Strict" guideline, where "Modern" is TLSv1.2 and TLSv1.3 without DHE suites, and "Modern-Strict" is TLSv1.3 only.  I think we should discuss this further and try to push our (outsider) recommendations into the security team at Mozilla.

janbrasna: Basically MTAs need different profiles for submission than for delivery etc., and even the current intermediate is not enough to support the reality in the wild… so we'll have to figure out a way how to support non-http server universe, while not restricting the development of the specs. One side would be the recommendations proper, that can add the rationale for different interop scenarios, but at the same time the config tool will need to surface that info too. TBD.

src/templates/partials/caddy.hbs
```diff
@@ -1,43 +1,31 @@
...
-    protocols tls1.0 tls1.3
-    {{#if (minver "2.0.0" form.serverVersion)}}
-    ciphers {{{join output.ciphers " "}}}
-    {{ else }}
-    ciphers ECDHE-ECDSA-AES128-GCM-SHA256 ECDHE-RSA-AES128-GCM-SHA256 ECDHE-ECDSA-AES256-GCM-SHA384 ECDHE-RSA-AES256-GCM-SHA384 ECDHE-ECDSA-WITH-CHACHA20-POLY1305 ECDHE-RSA-WITH-CHACHA20-POLY1305 ECDHE-ECDSA-AES128-CBC-SHA ECDHE-RSA-AES256-CBC-SHA ECDHE-RSA-AES128-CBC-SHA ECDHE-ECDSA-AES256-CBC-SHA RSA-AES128-CBC-SHA RSA-AES256-CBC-SHA RSA-3DES-EDE-CBC-SHA
-    {{/if}}
-}
-{{/if}}
-{{#if (includes "intermediate" form.config)}}
+example.com {
```
janbrasna, line 14: Do I read it correctly that the config was not scoping the settings to the hostname before? Is that some recent change, or a preference in style, or the configs were rendered completely broken?

gstrauss: was broooooken

janbrasna: (I know the TLSv flags were completely wrong etc., I was was working on a fix to both v1 and v2 before tracing back the v1 configs only to v5.4 specs and nothing more etc. some time ago… but never really noticed the whole config was missing the actual closure!…)

I may want to come back to this later, but if it's super broken, the same case… feel free to publish self-r+ and will do a proper review later, following up with any changes separately.

gstrauss: We should consider applying that logic to this whole site.  IMNSHO, we should be providing good advice for software released in the last 3 years, and should prioritize that much higher than pedantic historical curiosities for ancient configs.

I feel that there is no good reason to spend time updating the site for EOL software versions (caddy v1) or specs prior to guidelines 5.7 before the much higher priority of updating the site to provide good advice for contemporary software versions.

Per @janbrasna comment above, I will merge this as-is.

Please do feel free to revisit and improve.  Also, open invitation to other contributors, too!  @francislavoie, if nothing else, feedback will be more timely on PRs with @janbrasna and myself as new contributors to this github repo.

gstrauss:
rebased to tip of master and merged.

src/templates/partials/caddy.hbs
janbrasna, line 22: This should not filter on form inputs, but presence of values in output protocols; perhaps folded into the block above to cut repetition.

src/templates/partials/caddy.hbs
```diff
@@ -1,43 +1,31 @@
...
 
-# Due to a lack of DHE support, you -must- use an ECDSA cert to support IE 11 on Windows 7
-tls {
-    protocols tls1.2 tls1.3
-    {{#if (minver "2.0.0" form.serverVersion)}}
+  tls {
+{{#if output.ciphers.length}}
+    # Due to a lack of DHE support, you -must- use an ECDSA cert to support IE 11 on Windows 7
```
janbrasna, line 18: I will add a condition to show this only if there's IE11@W7 in the oldestClients, i.e. once the support profile shifts, we won't have to revisit similar hardcoded notes, and have it switched off in template logic.

src/templates/partials/caddy.hbs
```diff
@@ -1,43 +1,31 @@
...
-    ciphers ECDHE-ECDSA-AES128-GCM-SHA256 ECDHE-RSA-AES128-GCM-SHA256 ECDHE-ECDSA-AES256-GCM-SHA384 ECDHE-RSA-AES256-GCM-SHA384 ECDHE-ECDSA-WITH-CHACHA20-POLY1305 ECDHE-RSA-WITH-CHACHA20-POLY1305 ECDHE-ECDSA-AES128-CBC-SHA ECDHE-RSA-AES256-CBC-SHA ECDHE-RSA-AES128-CBC-SHA ECDHE-ECDSA-AES256-CBC-SHA RSA-AES128-CBC-SHA RSA-AES256-CBC-SHA RSA-3DES-EDE-CBC-SHA
-    {{/if}}
-}
-{{/if}}
-{{#if (includes "intermediate" form.config)}}
+example.com {
 
-# Due to a lack of DHE support, you -must- use an ECDSA cert to support IE 11 on Windows 7
-tls {
-    protocols tls1.2 tls1.3
```
janbrasna, line 25: I think I will add these commented out, combined with the "Caddy automatically configures safe TLS settings" note, to show what is the actual default, and that it just doesn't need setting.

src/templates/partials/caddy.hbs
```diff
@@ -1,43 +1,31 @@
 # {{output.header}}
 # {{{output.link}}}
-{{#unless (includes "old" form.config)}}
+{{#unless (minver "2.0.0" form.serverVersion)}}
+# note that Caddy version 1 reached end-of-life > 4 years ago back in 2020
+{{else}}
+{{#if (includes "old" form.config)}}
+# note that Caddy supports only TLSv1.2 and later
+{{else}}
```
janbrasna, line 8: Will need `includes "TLSv1.1" output.protocols` instead, and the note perhaps moved to the protocols line — making a note the old config can't be supported to that extent.

src/templates/partials/caddy.hbs
```diff
@@ -1,43 +1,31 @@
 # {{output.header}}
 # {{{output.link}}}
-{{#unless (includes "old" form.config)}}
+{{#unless (minver "2.0.0" form.serverVersion)}}
+# note that Caddy version 1 reached end-of-life > 4 years ago back in 2020
+{{else}}
```
janbrasna, line 5: I'd like to do that in a consistent way, as with unsupported tls13 — emit those empty configs similarly. More templates will need that, I'll follow up later.

janbrasna (commented):
Didn't review, let self-r+ as the current output was completely wrong so anything's an improvement.

Just from glancing I'd have some requests to address, but will come back to that later, review properly and follow up separately:

src/templates/partials/caddy.hbs
gstrauss: Please add to your wishlist to track there.

src/templates/partials/caddy.hbs
gstrauss: See #262

src/templates/partials/caddy.hbs
gstrauss: (todo; not done in #262)

src/templates/partials/caddy.hbs
gstrauss: See #262

src/templates/partials/caddy.hbs
gstrauss: See #262.  You might suggest additions to the comments.

src/templates/partials/caddy.hbs
janbrasna: This actually doesn't stop the output for Caddy 1, only adds a comment and then keep with Caddy 2 config that's not really compatible.

I'll look into all of the above and will co-author the followup PR as some of the changes needed won't be allowed to be submitted as review suggestions et al., so I'll just pick up over there…

PR #262, opened 2024-10-12
Replace config name with feature check in go/caddy conditions

gstrauss:
Replace `form.config` in template conditionals
Instead, prefer to check output.protocols in template conditionals.

### DIFF
--- a/src/templates/partials/caddy.hbs
+++ b/src/templates/partials/caddy.hbs
@@ -2,12 +2,6 @@
 # {{{output.link}}}
 {{#unless (minver "2.0.0" form.serverVersion)}}
 # note that Caddy version 1 reached end-of-life > 4 years ago back in 2020
-{{else}}
-{{#if (includes "old" form.config)}}
-# note that Caddy supports only TLSv1.2 and later
-{{else}}
-# note that Caddy automatically configures safe TLS settings
-{{/if}}
 {{/unless}}
 
 # replace example.com with your domain name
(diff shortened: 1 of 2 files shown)


gstrauss:
Aside: awselb.hbs references `PolicyName: Mozilla-{{form.config}}-v5-0`.  Should we open a separate issue to update this?  Do you happen to know what needs to happen to update the two references in that file to `Mozilla-{{form.config}}-v5-7`?

janbrasna:
For a good laugh here's what worked: https://github.com/janbrasna/ssl-config-generator/commit/9912704d6c2a13f115204d9732a068f35314b700 (so I may try looking into why sstls is not available on the output, rather than anything similar;D — or, ofc, exposing the `version` from `sstls` to `output` or via `form` if we want it there anyways…) — so yeah, noticed it before and tried to look for a quick fix, but will need arch decision first where else we might need it, and add it there properly.

gstrauss:
> For a good laugh here's what worked: [janbrasna@9912704](https://github.com/janbrasna/ssl-config-generator/commit/9912704d6c2a13f115204d9732a068f35314b700) (so I may try looking into why sstls is not available on the output, rather than anything similar;D — or, ofc, exposing the `version` from `sstls` to `output` or via `form` if we want it there anyways…) — so yeah, noticed it before and tried to look for a quick fix, but will need arch decision first where else we might need it, and add it there properly.

Not pretty code, but if it works, why not?  Short-term, I'd prefer to simply change "5.0" to "5.7" and revisit later.  Longer-term, I think I'd prefer having state.js use javascript to export a variable with the extracted value rather than the contortions in HandlebarJS templates.

In any case, awselb.hbs is not referencing `{{form.config}}` in a template *conditional*, so is not in scope for this PR.

Are the proposed code changes in this PR ready for merge?

gstrauss:
On Oct 13 in https://github.com/mozilla/ssl-config-generator/pull/262#issuecomment-2408871266 above, I wrote:
> Are the proposed code changes in this PR ready for merge?

You marked the comment as Resolved.

A month has passed.  I am merging this PR.  @janbrasna this is an example of you not being able to reliably make progress on small, well-scoped changes (or to properly flag the PR with data why the PR is blocked).   Note: I created this PR specifically to address *your* feedback in another PR.

PR #263, opened 2024-10-12
Enable OpenSSL 3.x as the default in UI

gstrauss:
In #256 the support for v3 changes is being added. It can be released without making the 3.x branch the default for everyone, allowing some time to offer it as a preview feature, as "opt-in", before settling on that as the new default for anyone and everyone.

This is a followup, to enable the v3.x as our new default, to stage the rollout in two phases — once we're happy with the #256 changes for the current 1.1.1 configs (e.g. where changing dhparams for the current defaults), as well as any feedback on the new changes (leaving out dhparams in favour of handshake automagic selection, and lowering seclevel for `old` via patching the cipherstring) it brings once switched as the new default.

### DIFF
--- a/src/js/configs.js
+++ b/src/js/configs.js
@@ -118,7 +118,7 @@ module.exports = {
     tls13: '1.13.0',
   },
   openssl: {
-    latestVersion: '1.1.1w',
+    latestVersion: '3.4.0',
     tls13: '1.1.1',
   },
   oraclehttp: {


gstrauss:
Note, if any profiles recommend DHE ciphers, then should add a comment about configuring key size limits to avoid resource attacks which negotiate very large key sizes?  See #162

src/js/configs.js
```diff
@@ -118,7 +118,7 @@ module.exports = {
     tls13: '1.13.0',
   },
   openssl: {
-    latestVersion: '1.1.1w',
+    latestVersion: '3.3.2',
```
janbrasna, line 121: The 3.4.0 should now include additional DHEattack CVEs fixed, FYI:

```suggestion
    latestVersion: '3.4.0',
```

janbrasna (changes requested):
The biggest impact is perhaps that OpenSSL 3.x adds support for FFDH exchange in TLSv1.3, basically bringing back the never-ending performance/DoS issue #162 officially also to `modern` (and other configs even when we remove DHE-RSA suites).

See https://openssl-library.org/post/2022-10-21-tls-groups-configuration/

Before releasing this as the new default, we should either consider configuring bit size limits (=restrict RFC7919 groups) to avoid malicious clients negotiating unreasonably large key exchanges, or basically just update the upstream wording from "curves" to "groups" to only include EC curves in RFC7919/RFC8446 negotiations, and rule out any `ffdhe*` handshakes.

gstrauss:
> Before releasing this as the new default, we should either consider configuring bit size limits (=restrict RFC7919 groups) to avoid malicious clients negotiating unreasonably large key exchanges, or basically just update the upstream wording from "curves" to "groups" to only include EC curves in RFC7919/RFC8446 negotiations, and rule out any ffdhe* handshakes.

That is being discussed in #270 Explicitly configure curves/groups from the guidelines

---

The issue I have with making #270 a blocking issue for this is that people are already using OpenSSL 3.x, as all versions of OpenSSL prior to OpenSSL 3.0 are now EOL.  Yes, some people are still running older versions of OpenSSL, but anyone maintaining their software is not running older versions.

gstrauss:
Included in #281

PR #283, opened 2024-12-02
use js to replace jquery and bootstrap; keep css

gstrauss:
*  use js to replace jquery and bootstrap; keep css
*  retain css from bootstrap, generated and pruned as src/css/bs-4.6.2.css
*  remove bootstrap, jquery, popper; upgrade rest
    
    note: upgrading semver from 6.3.1 to 7.6.3 breaks minver() comparision
    of openssl versions, so semver held back at 6.3.1 where, for example,
    openssl 1.1.1w > 1.1.1 (as expected)
    
    note: extract-loader kept at 4.0.3 instead of 5.1.0 since 'npm audit'
    reports security issues in dependencies for extract-loader 5.1.0

---

This PR uses plain javascript, widely supported on modern browsers, to implement the basic functionality of the site.

There are some minor UX changes:
* Checkboxes under Miscellaneous config are hidden rather than grayed out when the app does not support them.
* "Copied!" message when Copy button is pressed is right below the button rather than a popup.
* updated highlight.js package uses different visual colors for highlighting

Subsequent patches are welcome if you feel that the UX needs tweaks.


gstrauss:
Does highlighter.js make a worthwhile improvement to the ssl-config-generator output?  It adds about 130KB to the page, and ten (10) of the apps listed in `src/js/configs.js` use 'nginx' as highlighter since there is not a matching config highlighter.

May I remove highlighter and see if anyone comments about it missing?

gstrauss:
If anyone with UX and javascript expertise has other ideas for the GUI, please consider opening a PR with suggestions.  Thank you.

gstrauss:
This was reverted and pieces have subsequently been committed in #288 and #291

The jquery dependency remains, for the moment.

Issue #270, opened 2024-11-11
Explicitly configure curves/groups from the guidelines


janbrasna:
In the past this was part of the configs, only allowing curves enumerated in the json data for the configs. At some point it got removed due to bugs in implementations, and also clients and servers being able to negotiate the reasonable thing anyways.

### History

The original issue #44 when lighttpd wouldn't support multiple values is hopefully just a thing of the past. The support in nginx template was removed (or not added at all) due to reasons described in #76 (for details see https://trac.nginx.org/nginx/ticket/1089 — e.g. this needs #141 resolved first).

### Background

If we want to officially support OpenSSL 3.x in #260, the issue with performance #162 is relevant again, this time even for `modern` (and way past any TLSv1.2 DHE suites removal from the lists) — as OpenSSL 3.x started supporting FFDH key exchange for TLSv1.3 — which is part of RFC 8446 requiring RFC 7919, allowing to negotiate unnecessarily large keys (see https://openssl-library.org/post/2022-10-21-tls-groups-configuration/ for mitigation recommendations from the authors).

See the overview at https://datatracker.ietf.org/doc/html/rfc7919#section-2 for addition of groups to curves, and their codepoints.

This wasn't an issue with OpenSSL 1.x.x as it only had support for ECDH key exchange in TLSv1.3, so we didn't have to restrict anything really. That changes with OpenSSL 3.x adding support for FFDH in TLSv1.3 (and others, too: see https://dheatattack.gitlab.io/mitigations/#tls for e.g. GnuTLS versions.)

BTW for the future of OpenSSL 3.x support for RFC 7919 also in TLSv1.2 follow:
- https://github.com/openssl/openssl/issues/10971
- https://github.com/openssl/openssl/pull/24551

### Actions

To officially extend the support to the current OpenSSL versions that now allow (and enable by default without any group restriction) FFDH too, and keep the same guidelines, we have to consider either explicitly restricting the RFC 7919 groups to a reasonable subset (as some servers already do), or decide on not allowing any FFDH key exchange to begin with, and resort back to just recommending EC curves (as we currently do, just neither explain it, not enforce it though).

This may need revisiting all the "curves" settings in servers, whether they mean any RFC 7919 groups by that (as most would probably silently do), or if they separate ECDH and FFDH settings and configure both independently (as some do, with good defaults FWIW) — to give us a better idea of how such approach is feasible, and whether we'll be able to limit or completely restrict the groups allowed.

[...]

PR #284, opened 2024-12-03
config TLSv1.3 key exchange groups where supported

gstrauss:
config TLSv1.3 key exchange groups where supported

The reason to explicitly configure Groups is due to OpenSSL 3.0 adding ffdhe* groups to the defaults, potentially exposing servers supporting TLSv1.3 and OpenSSL 3.0 to DHEater attacks. (See https://github.com/mozilla/ssl-config-generator/issues/162)

Note: .hbs additions have been made from the documentation and have not been tested on the all the server apps.
Some slight syntax adjustment may turn out to be needed.

Note: This PR exposes the TLS curves that are already defined in the guidelines.  This is not a change to the guidelines.  However, the curves were not previously exposed in ssl-config-generator configs.  This PR implements the Groups/Curves config in server apps that have config support.

### DIFF
--- a/src/js/state.js
+++ b/src/js/state.js
@@ -97,6 +97,7 @@ export default async function () {
       supportsConfigs: configs[server].supportsConfigs !== false,
       supportsHsts: configs[server].supportsHsts !== false,
       supportsOcspStapling: supportsOcspStapling,
+      tlsCurves: ssc.tls_curves,
       usesDhe: ciphers.join(":").includes(":DHE") || ciphers.join(":").includes("_DHE_"), 
       usesOpenssl: configs[server].usesOpenssl !== false,
     },
--- a/src/templates/partials/apache.hbs
+++ b/src/templates/partials/apache.hbs
@@ -39,6 +39,9 @@
 
 # {{form.config}} configuration
 SSLProtocol             {{{join output.protocols " "}}}
+{{#if (minver "2.4.8" form.serverVersion)}}
+SSLOpenSSLConfCmd       Curves {{{join output.tlsCurves ":"}}}
+{{/if}}
 {{#if output.ciphers.length}}
 SSLCipherSuite          {{{join output.ciphers ":"}}}
 {{/if}}
--- a/src/templates/partials/caddy.hbs
+++ b/src/templates/partials/caddy.hbs
@@ -14,9 +14,10 @@ example.com {
 
   {{/if}}
     # Note: Caddy automatically configures safe TLS settings,
-    # so 'ciphers' may safely be commented out to use Caddy defaults.
+    # so 'ciphers' and 'curves' may safely be commented out to use Caddy defaults.
     ciphers {{{join output.ciphers " "}}}
 {{/if}}
+    curves x25519 secp256r1 secp384r1
 {{#if (includes "TLSv1.2" output.protocols)}}
   {{#if (includes "TLSv1.1" output.protocols)}}
     # Note: Caddy supports only TLSv1.2 and later
--- a/src/templates/partials/dovecot.hbs
+++ b/src/templates/partials/dovecot.hbs
@@ -20,6 +20,7 @@ ssl_min_protocol = {{output.protocols.[0]}}
 {{else}}
 ssl_protocols = {{join output.protocols " "}}
 {{/if}}
+ssl_curve_list = {{{join output.tlsCurves ":"}}}
 {{#if output.ciphers.length}}
 ssl_cipher_list = {{{join output.ciphers ":"}}}
 {{/if}}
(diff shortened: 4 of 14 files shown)


gstrauss:
Repeating from my https://github.com/mozilla/ssl-config-generator/issues/270#issuecomment-2498615793:

IANA TLS Supported Groups (see "Recommended" column in the table)
https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8

The recommended groups are currently X25519:P-256:P-384:X448.
(P-256 is also known as secp256r1 and as prime256v1)
(P-384 is also known as secp384r1)
Note: different TLS libraries may recognize some of the string aliases, but might not recognize all of the string aliases.
Aside: OpenSSL, GnuTLS, mbedTLS, WolfSSL support X448; whereas NSS, BoringSSL, LibreSSL do not support X448

gstrauss:
Almost half of the servers in ssl-config-generator do not support configuring a list of Groups/Curves in their config files:

* no config for Groups/Curves
  AWSALB, AWSELB, MySQL, OracleHTTP, Redis

* likely configurable in XML (or on command line) depending on underlying java TLS provider
  Jetty, Tomcat 

* config option for single curve (recommend: prime256v1, but Squid has additional limitations)
  Coturn, Squid

Of note, configuring a curve in Squid disables DHE key exchange.  It is one or the other, and the default is DHE, not EC.  This is a serious limitation.  If someone files a bug upstream with Squid, please reference here.

gstrauss:
Some servers added support to configure Groups only in recent versions.  Select examples (incomplete list):
* postgresql supports a single curve in `ssl_ecdh_curve` before postgresql 18.0.0
  postgresql 18.0.0 supports a list of curves in `ssl_groups`
  All Postresql versions use "prime256v1" as the default if `ssl_ecdh_curve` or `ssl_groups` is not specified
* exim 4.86+ supports a single curve in `tls_eccurve`, and a list of curves is supported in exim 4.97
  https://bugs.exim.org/show_bug.cgi?id=2955

gstrauss:
boringssl and libressl do not support ffdhe key exchange with TLSv1.3.
This fails: `openssl s_client -tls1_3 -groups ffdhe2048 -connect google.com:443`
The same command tested on my local machine (changing google.com:443 to localhost:443) works with openssl 3.2.2, but again, does not work with boringssl or libressl.

gstrauss:
Aside: some notes for configuring TLS in Java, specifically `jdk.tls.namedGroups`,
Note: not implemented in ssl-config-generator for Jetty or Tomcat

Java
https://www.java.com/en/configure_crypto.html
Disabling weak named curves by default in TLS, CertPath, and Signed JARs
...
Curves that will remain enabled are: secp256r1, secp384r1, secp521r1, X25519, and X448.
...
Adding support for the x25519 and x448 named elliptic curve groups to TLS
The named elliptic curve groups x25519 and x448 will be available for JSSE key agreement in TLS versions 1.0 to 1.3, with x25519 being the most preferred of the default enabled named groups. After this change, the default ordered list will be:
x25519, secp256r1, secp384r1, secp521r1, x448, ffdhe2048, ffdhe3072, ffdhe4096,
ffdhe6144, ffdhe8192
The default list can be overridden using the system property jdk.tls.namedGroups.

https://docs.oracle.com/en/java/javase/23/security/transport-layer-security-tls-protocol-overview.html#GUID-C6554A00-CF26-4661-991D-EA1B9EA6CCBE
https://docs.oracle.com/en/java/javase/23/security/java-secure-socket-extension-jsse-reference-guide.html#GUID-A41282C3-19A3-400A-A40F-86F4DA22ABA9


Tomcat also has option to use tcnative (Tomcat Native) library to use OpenSSL as a replacement for JSSE to support TLS connections.  https://tomcat.apache.org/native-doc/

gstrauss:
Golang has different names for Groups than OpenSSL.  Instead of providing a mapping, this PR hard-codes the list of Groups in `caddy.hbs`, `go.hbs`, `traefik.hbs`.  Caddy documents a different spelling from Go and Traefik for Groups permitted in Caddy config `curves`: https://caddyserver.com/docs/caddyfile/directives/tls

gstrauss:
* openssl 1.1.1 TLSv1.3 support
* openssl 3.0.0 TLSv1.3 support with DHE key exchange using ffdhe* Groups
              (ffdhe* Groups added to openssl ssl/t1_lib.c:supported_groups_default[]
               (renamed from ssl/t1_lib.c:eccurves_default[]))
              https://github.com/openssl/openssl/commit/9aaecbfc98eb89a03f72b35d343e08f377e7803a
              https://github.com/openssl/openssl/pull/8178
* Note: configuring Groups also affects TLSv1.2 kECDHE ciphers

PR #285, opened 2024-12-05
Revert upgrades in #283

gstrauss:
Revert upgrades in #283.  What worked with my local `npm run watch` has issues in production.


gstrauss:
Among other things, some more work is needed around the Copy button feedback.
```
Source: SSLConfigGenerator.server_change() 22 [ssl-config.mozilla.org](https://ssl-config.mozilla.org/)
Content-Security-Policy: The page’s settings blocked an inline style (style-src-attr) from being applied because it violates the following directive: “style-src 'self' https://code.cdn.mozilla.net/”
Source: display:none
```

PR #326, opened 2025-02-18
go: fix unescaped newline in generated string literal

dusanbrankov:
fix syntax error in the generated Go code and ensure newline in HTTP reply


gstrauss:
Thank you!

Issue #323, opened 2025-01-27
Remove OCSP recommendation


Gunni:
OCSP is being sunset for Let's Encrypt this year (2025).

https://letsencrypt.org/2024/12/05/ending-ocsp/

Generated config should not add OCSP config.

> - January 30, 2025
>   - OCSP Must-Staple requests will fail, unless the requesting account has previously issued a certificate containing the OCSP Must Staple extension
> - May 7, 2025
>   - Prior to this date we will have added CRL URLs to certificates
>   - On this date we will drop OCSP URLs from certificates
>   - On this date all requests including the OCSP Must Staple extension will fail
> - August 6, 2025
>   - On this date we will turn off our OCSP responders

gstrauss:
Thank you for the note.

OCSP support will probably not be removed from the SSL config generator, though we may consider unchecking the option box by default rather than the current default of enabling OCSP in the configs.

PR #331, opened 2025-03-05
refresh Server Side TLS guidelines: proposal for v6.0

gstrauss:
The existing Mozilla Server Side TLS guidelines are from 2020 (with one small change since) and are long overdue for review and update.

As a member of Mozilla SSL Config Generator working group, I have compiled the following proposal for a new version of the Mozilla Server Side TLS guidelines.  My intent in this PR is to update the guidelines in line with existing recommendations from CDNs and other large tech companies; I do not believe that any of the changes are ground-breaking.

This proposal is being filed as a PR to the ssl-config-generator repository in order to have the changes available as git diffs for review.  Please review the commits in this PR individually and sequentially to visualize the changes.

To set some expectations: this PR is being submitted for review and comment on the proposed changes herein.  If there are other items that knowledgable parties would like to suggest changing, please file them as separate issues at 
https://github.com/mozilla/server-side-tls/issues and they will be considered there for this or a future update to the guidelines.

Thank you!

---

guideline 5.8 proposed changes

* remove kDHE ciphers from Intermediate and Old
* change Old dhParamSize from 1024 to 2048 and
  use ffdhe2048 instead of locally generated dhparams
  * DHE ciphers should be disabled in Intermediate compatibility configuration
     https://github.com/mozilla/server-side-tls/issues/268
  * Stop recommending DHE, because of "dheater" vulnerability
     https://github.com/mozilla/server-side-tls/issues/299

* meta: recommended ciphers for Intermediate are now all PFS with AEAD.
  Today, in 2025, this is widely supported on the internet.
  Impact: no longer support IE11 on Windows 7 in Intermediate and Modern configs.
  Today, in 2025, IE11 and Windows 7 should be omitted from the the public
  Mozilla Server Side TLS guidelines for Intermediate or Modern configs.

ssl-config-generator rendering/behavior changes:

* default to uncheck box for OCSP stapling, rather than default checked
  * Let's Encrypt: Intent to End OCSP Service
     https://letsencrypt.org/2024/07/23/replacing-ocsp-with-crls/
  * Let's Encrypt: Ending OCSP Support in 2025
     https://letsencrypt.org/2024/12/05/ending-ocsp/
  * Remove OCSP recommendation
     https://github.com/mozilla/ssl-config-generator/issues/323
* For HSTS, use HTTP status code 308 Permanent Redirect for redirecting all http:// to https://
  so that HTTP methods which are not GET or HEAD are redirected without method change to GET
  * The Hypertext Transfer Protocol Status Code 308 (Permanent Redirect)
     https://www.rfc-editor.org/rfc/rfc7238 (published June 2014)
  * 308 Permanent Redirect
     https://github.com/mozilla/ssl-config-generator/issues/117

guideline 6.0 proposed changes

* remove "Old" configuration; "Old" configuration is not recommended
  For those who need to reference older guidance, older versions of
  the ssl-config-generator may be obtained from the Internet Archive
  https://web.archive.org/web/*/https://mozilla.github.io/server-side-tls/ssl-config-generator/
  Alternatively, the current ssl-config-generator will continue to render URLs which specify older guideline versions (e.g. `&guideline=5.7`) and will contain the recommendations in the older guideline (e.g. cipher lists), but the *rendering* may not exactly match what was produced years ago since the ssl-config-generator presentation may have changed in the interim.


ghen2:
Please also take into account PQC key exchange (X25519MLKEM768) as supported by modern browsers.
OpenSSL 3.5.0 will be released in April 2025 with support for PQC key exchange enabled by default.

gstrauss:
> Please also take into account PQC key exchange (X25519MLKEM768) as supported by modern browsers. OpenSSL 3.5.0 will be released in April 2025 with support for PQC key exchange enabled by default.

Agree.  Should be considered for Modern config, when supported by TLS libraries.

ghen2:
> Agree. Should be considered for Modern config, when supported by TLS libraries.

Why only Modern?  All profiles have the same level of forward compatibility, only different backwards compatibility?

tomato42:
Yes, we should definitely include PQC in the TLS groups (curves) listing. in all security levels.

I would say that the X25519MLKEM768 is absolute minimum, Secp256r1MLKEM768 *should* be included too (will be supported, or already is supported by Firefox), I do wonder about Secp384r1MLKEM1024... that one is a bit further away with support in Firefox, otoh, we do include secp384r1 in the curves already, so including PQC hybrid variant of it in the listing would be consistent...

ghen2:
In light of https://github.com/lamps-wg/draft-composite-kem/issues/123 I wouldn't necessarily support all possible combinations.
We can turn the question around: will any client/browser _not_ be implementing X25519MLKEM768 ?

(a backup KEM can be added once standardized, eg. based on HQC, but additional ML-KEM hybrid's aren't backups)

tomato42:
> (a backup KEM can be added once standardized, eg. based on HQC, but additional ML-KEM hybrid's aren't backups)

the others are there to meet operational requirements, not interoperability requirements, there very likely will be environments which cannot use x25519, or are required to use ML-KEM-1024

ounsworth:
> In light of [lamps-wg/draft-composite-kem#123](https://github.com/lamps-wg/draft-composite-kem/issues/123) I wouldn't necessarily support all possible combinations. We can turn the question around: will any client/browser _not_ be implementing X25519MLKEM768 ?
> 
> (a backup KEM can be added once standardized, eg. based on HQC, but additional ML-KEM hybrid's aren't backups)

Sorry to be late to this thread.

As @tomato42 alludes to, only ML-KEM-1024 is allowed by CNSA 2.0 and the Australian equivalent.

Also, X25519 is not universally-allowed, so many applications will require a P-256 variant.

I would suggest that you want two Mandatory-to-implements:

* MLKEM768+X25519
* MLKEM768+P256

and RECOMMENDED support for MLKEM1024+P384.

gstrauss:
OpenSSL 3.5 LTS was released April 2025 with support for PQC hybrid KEMs.
OpenSSL 4.0 LTS is scheduled to be released April 2026.
After Oct 22, 2026, no earlier versions of OpenSSL will be publicly supported by the OpenSSL Foundation.
(excluding paid premium support)

gstrauss:
I updated the proposal for 5.8 to add only X25519MLKEM768, and, for now, omit "SecP256r1MLKEM768".
```
-            "tls_curves": ["X25519", "prime256v1", "secp384r1"],
+            "tls_curves": ["X25519MLKEM768", "X25519", "prime256v1", "secp384r1"],
```
"SecP256r1MLKEM768" may be reconsidered in a future update.

nuxi:
I am excited to see that the Mozilla TLS guide is getting some updates, so thank you very much. I saw your comment on https://github.com/mozilla/server-side-tls/issues/304 and so I figured I would give my $0.02.

The new intermediate standard is pretty much exactly what I use at work. The only comment I would have made is already addressed but I will also give a :+1: to ommitting `SecP256r1MLKEM768` for now.

While `SecP256r1MLKEM768` is implemented in many TLS libraries, its not actually in use by web browsers. In the past, this lack of adoption by web browsers has been considered adequate reason to omit something from the guidelines. Even amongst TLS libraries, I do not believe any of them send a `SecP256r1MLKEM768` key share by default. This can always be reconsidered later if things change.

IMHO those who need to support special requirements like FIPS or Common Criteria should be prepared to tweak the guidelines as needed to conform to those special requirements. (Unfortunately for me, I am one of those people :cry:)

PR #333, opened 2025-03-21
Add support for OpenLiteSpeed

Code-Egg:


Code-Egg:
Hi @gstrauss , would you be able to review this pull request whenever it's convenient for you?

gstrauss:
@Code-Egg Thank you for the submission.  I will take a closer look this coming weekend.  Apologies for the delayed response.

Does OpenLiteSpeed really require magic numbers for `sslProtocol`?

Is it easy to add configuration to redirect all http requests to https when HSTS is enabled?

Code-Egg:
Thanks for the fixes. 

About What happend to # Virtual Host Level Configuration and vhssl? Should the comment be removed?
Either way is fine. Since the HTTP Strict Transport Security (HSTS) header can only be set at the virtual host level, the comment mainly serves as a reminder.

